PR #1, opened 2018-12-23
Update rocker to use entry points for plugins

tfoote:
As well as improve several other aspects

potential 18.04 support
adding parameters
cleaner outputs and error checking


tfoote:
validated with iris and plane demos

PR #5, opened 2019-01-03
Add support for mounting devices

tfoote:
For example to be able to pass through /dev/input/js0


tfoote:
Validated against car_demo with joystick driving.

Issue #14, opened 2019-01-10
nvidia overlay failing on bionic: x86_64-linux-gnu: no such file or directory


ruffsl:
The auto generated Dockerfile seems to encounter a build issue in the first stage of it multistage build.

``` shell
$ rocker --nvidia --user --pull --pulse osrf/ros:crystal-desktop rviz2
Plugins found: ['dev_helpers', 'home', 'nvidia', 'pulse', 'user']
Active extensions ['nvidia', 'pulse', 'user']
Pulling image osrf/ros:crystal-desktop
b'{"status":"Pulling from osrf/ros","id":"crystal-desktop"}\r\n'
b'{"status":"Digest: sha256:62579d806386b2b3ff2ecf3924f0ba39b8d8422fb7979db42e469cfbfd26e19f"}\r\n'
b'{"status":"Status: Image is up to date for osrf/ros:crystal-desktop"}\r\n'
Writing dockerfile to /tmp/tmpg_lv8g3_/Dockerfile
vvvvvv
...
^^^^^^
Building docker file with arguments:  {'path': '/tmp/tmpg_lv8g3_', 'rm': True, 'decode': True, 'nocache': False, 'tag': 'rocker_osrf/ros:crystal-desktop_nvidia_pulse_user'}
building > Step 1/16 : FROM nvidia/opengl:1.0-glvnd-devel-ubuntu18.04 as glvnd
building >  ---> 04c47f164887
building > Step 2/16 : FROM osrf/ros:crystal-desktop
building >  ---> c3d97d8c2f50
building > Step 3/16 : COPY --from=glvnd /usr/local/lib/x86_64-linux-gnu /usr/local/lib/x86_64-linux-gnu
Cannot run if build has not passed.
```

Would be nice if the rocker could forward the build error to the printout, rather than needing to manually build the same dockerfile to identify the exact error.

``` shell
$ docker build -t temp /tmp/tmpg_lv8g3_/
Sending build context to Docker daemon  4.096kB
Step 1/16 : FROM nvidia/opengl:1.0-glvnd-devel-ubuntu18.04 as glvnd
 ---> 04c47f164887
Step 2/16 : FROM osrf/ros:crystal-desktop
 ---> c3d97d8c2f50
Step 3/16 : COPY --from=glvnd /usr/local/lib/x86_64-linux-gnu /usr/local/lib/x86_64-linux-gnu
[...]

PR #15, opened 2019-01-10
Nvidia 18.04 support

tfoote:
This is changing the nvidia snippet to be conditioned on 16.04 vs 18.04.
Following behaviors from: 
- https://gitlab.com/nvidia/opengl/blob/ubuntu16.04/glvnd/runtime/Dockerfile
vs
- https://gitlab.com/nvidia/opengl/blob/ubuntu18.04/glvnd/runtime/Dockerfile

Currently it's conditioned on the host OS, however because the snippet is done in the image I think it would be better to have it conditioned on the image's OS. 

I'd originally thought that it was coupled to the host OS as you need the same nvidia drivers as the host. And it worked for 16.04 host and an 18.04 image using the 16.04 injection. However when using the 18.04 host it broke in #14. 

@ruffsl can you verify that this works for you? And we can figure out how the parameterization should be done?

### DIFF
--- a/README.md
+++ b/README.md
@@ -27,11 +27,6 @@ Debian packages are available from the ROS repositories. You can set them up in 
 
 Then you can `sudo apt-get install python3-rocker`
 
-On Ubuntu older than bionic you will need to install python3-distro manually first. 
-
-See https://github.com/osrf/rocker/blob/master/backport_xenial.bash for tested on xenial.
-
-
 ## PIP
 
 Rocker is available via pip you can install it via pip using
--- a/README.md
+++ b/README.md
@@ -10,14 +10,18 @@ Rocker supports extensions via entry points there are some built in but you can 
 
 # Prerequisites
 
-This should work on most systems using with a recent docker version available. 
+This should work on most systems using with a recent docker version available.
+
+Docker installation instructions: https://docs.docker.com/install/
 
 ## NVIDIA settings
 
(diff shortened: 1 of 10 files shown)


ruffsl:
I had to update my python docker module in the env environment, but other than that it seems to be working. I can't seem to kill/exit with crtl+c though.

tfoote:
@ruffsl I rebased this on top of #24 could you validate it works for you on 18.04?

ruffsl:
Awsome, its working rather slick with rviz2 on my nvidia laptop. However, do we still need the `FROM nvidia` to `COPY` the `10_nvidia.json` if its >= 18.04? Kind of a bummer I wait to download ~2.47GB on my home network for 107 bytes of json. @flx42 does this file change often? Is there tracked URL we could fetch it?

``` Dockerfile
# Preamble from extension [nvidia]
# Ubuntu 16.04 with nvidia-docker2 beta opengl support
FROM nvidia/opengl:1.0-glvnd-devel-ubuntu18.04 as glvnd

FROM osrf/ros:crystal-desktop
# Snippet from extension [nvidia]
RUN apt-get update && apt-get install -y --no-install-recommends \
    libglvnd0\
	libgl1 \
	libglx0 \
	libegl1 \
	libgles2 \
    && rm -rf /var/lib/apt/lists/*
COPY --from=glvnd /usr/share/glvnd/egl_vendor.d/10_nvidia.json /usr/share/glvnd/egl_vendor.d/10_nvidia.json

ENV NVIDIA_DRIVER_CAPABILITIES ${NVIDIA_DRIVER_CAPABILITIES:+$NVIDIA_DRIVER_CAPABILITIES,}graphics
```

PR #18, opened 2019-01-16
Install either docker or docker-py

### LINKED ISSUE #17: Use docker instead of docker-py
The `docker-py` library was renamed to `docker` in 2.0.0 https://github.com/docker/docker-py/releases/tag/2.0.0 . If both `docker-py` and `docker` are installed then `import docker` is the module that was found first.

sloretz:
Fixes #17 

On `xenial` I ran `python setup.py install`:
* with `docker-py==1.8.0` installed
   * pip freeze shows only `docker-py` is installed
   * `rocker fedora bash` and `rocker --pull fedora bash` worked correctly
* with `docker` installed
   * pip freeze shows only `docker` is installed
   * `rocker fedora bash` and `rocker --pull fedora bash` worked correctly
* with nothing installed
   * pip freeze shows `docker-py` was installed

Have NOT tested:
* bionic

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -1,13 +1,33 @@
 #!/usr/bin/env python3
 
+import os
 from setuptools import setup
 
 install_requires = [
     'distro',
-    'docker-py',
     'empy',
     'pexpect',
 ]
+
+# docker API used to be in a package called `docker-py` before the 2.0 release
+docker_package = 'docker'
+try:
+    import docker
+except ImportError:
+    # Docker is not yet installed, pick library based on platform
+    # Use old name if platform has pre-2.0 version
+    if os.path.isfile('/etc/os-release'):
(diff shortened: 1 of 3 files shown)


tfoote:
This looks good. We can probably remove a bunch of this backport logic too: https://github.com/osrf/rocker/blob/master/backport_xenial.bash I'll see what I can clear from that too if we can have backwards compatibility with the pre 2.0 docker.

sloretz:
@tfoote OK to merge this one?

tfoote:
I merged this with a rebase: f787ae8a6d34e132284fe85258c8aa27bc2d0649 and 8433976bc39f811100c2e56a2b0bdb6c3c1523b7

 simplified the backports in 8246052faaff293fca285ca40e2edfbf8ec62041

PR #41, opened 2019-03-13
Save Dockerfile and run-docker.sh to a persistent location

wecacuee:
Add ability to save Dockerfile and run-docker.sh to a persistent location so that rocker can be used a boiler-plate Dockerfile generator.

### DIFF
--- a/src/rocker/cli.py
+++ b/src/rocker/cli.py
@@ -36,6 +36,8 @@ def main():
     parser.add_argument('--pull', action='store_true')
     parser.add_argument('--network', choices=['bridge', 'host', 'overlay', 'none'])
     parser.add_argument('--devices', nargs='*')
+    parser.add_argument('--docker-dir')
+    parser.add_argument('--run_docker_script_path', default="run-docker.sh")
 
     plugins = list_plugins()
     print("Plugins found: %s" % [p.get_name() for p in plugins.values()])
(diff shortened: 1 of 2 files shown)


Xia0ben:
Hello there,

Nice tool ! I was think of coding something like that and would need the extra capability offered by this pull request : is there any way to help bring it into the main branch ?

Have a nice day !

tfoote:
I'd like to understand the use case and proposed workflow a little better to make sure that this is the right format and we're not boxing ourselves in for future extensions in this direction. For example the whole thing could be written into a shell script including the dockerfile as a heredoc potentially. If there's a case for having a generator should we more formally restructure things to support that better?

Also if we're to do this it shouldn't write the files unless requested.

wecacuee:
@Xia0ben I stopped using this feature myself. After I wrote this, I realized that I was using rocker wrong. `rocker` is supposed to used more as an alternative to `docker run ...` then `docker build ...`. The Dockerfiles generated by rocker are machine and user specific hence should not be persisted or shared. 
May be your workflow is different. My workflow does not need this feature anymore. Although I would love to have additional `--run-args` as a feature. I would open a new request for that.

tfoote:
Thanks for the followup @wecacuee that makes sense about how you're looking for something slightly out of scope of what rocker is designed for. 

Suggestions for more options support sound good. I'll close this for now but please open a new issue or PR with your suggestions.

Issue #45, opened 2019-06-08
Add a flag to pass docker parameters


felixvd:
It would be helpful if it was possible to pass parameters with which to call `docker run`. In the case of the MoveIt docker image, the rocker `--user` flag does not work, but a custom command would have made a fine workaround. This has come up in https://github.com/ros-planning/moveit.ros.org/pull/282

I'm not clear on best practices here, but naively I would have imagined a call like this:

`rocker --nvidia --x11 --docker-run-options '-v /home/my_user/ws_moveit:/home/root/ws_moveit' moveit/moveit:master-source bash`

Might be related to #19; I'm not sure I understand the scope of that.

tfoote:
Extensions can support taking arbitrary arguments. @sloretz already has an arbitrary mount extension here: https://github.com/sloretz/off-your-rocker which does exactly what you mention. I'd rather keep to the programmatic API than just start passing through arbitrary commands. Though maybe an extension that does exactly that might make sense for maximum flexibility.

Your current situation might be complicated by #42 

@sloretz What do you think about promoting `oyr-mount` to being here as `mount`?

sloretz:
I'd be happy to open a PR for `mount` here. I also have a branch with a flag for passing any run arg, but I haven't released it yet https://github.com/sloretz/off-your-rocker/pull/1/files

tfoote:
Let's start with mount here. And we can leave the arbitrary parameters in oyr for now.

PR #46, opened 2019-06-20
Add --mount

sloretz:
This adds `--mount` which adds paths to a container using docker volumes.

Context https://github.com/osrf/rocker/issues/45#issuecomment-501518581 .


src/rocker/mount_extension.py
```diff
@@ -0,0 +1,52 @@
...
+        return ''
+
+    def get_snippet(self, cli_args):
+        return ''
+
+    def get_docker_args(self, cli_args):
+        args = ['']
+        for mount in cli_args['mount']:
+            mount = os.path.abspath(mount)
+            args.append('-v {0}:{0}'.format(mount))
```
tfoote: Instead of forcing it to be in the same place on the target, could we just pass through the argument directly. Then we can rely on the default docker parsing of the volume argument.  Otherwise we're basically preventing the user from doing more expressive interactions. 
```suggestion
            args.append('-v {0}'.format(mount))
```

src/rocker/mount_extension.py
```diff
@@ -0,0 +1,52 @@
...
+        return ' '.join(args)
+
+    @staticmethod
+    def register_arguments(parser):
+        parser.add_argument('--mount',
```
tfoote: If we pass it through we can just call this the volume option. Especially since there's actually a more expressive `--mount` option now: https://docs.docker.com/storage/bind-mounts/

README.md
```diff
@@ -120,4 +120,12 @@ On Xenial
...
 
-    rocker --nvidia osrf/ros:melodic-desktop-full gazebo
\ No newline at end of file
+    rocker --nvidia osrf/ros:melodic-desktop-full gazebo
+
+## Mount volumes
+
+`--mount` adds paths as docker volumes.
+The path used inside the container is the same as the path outside.
+The last path must be terminated with two slashes `--`.
```
tfoote:
```suggestion
The last path must be terminated with two dashes `--`.
```

README.md
```diff
@@ -120,4 +120,12 @@ On Xenial
...
 On Bionic
 
-    rocker --nvidia osrf/ros:melodic-desktop-full gazebo
\ No newline at end of file
+    rocker --nvidia osrf/ros:melodic-desktop-full gazebo
+
+## Mount volumes
+
+`--mount` adds paths as docker volumes.
+The path used inside the container is the same as the path outside.
```
tfoote: This limitation would go away if we support the longer syntax.

README.md
```diff
@@ -120,4 +120,12 @@ On Xenial
...
+The last path must be terminated with two slashes `--`.
+
+    rocker --mount ~/.vimrc ~/.bashrc -- ubuntu:18.04
```
tfoote:
```suggestion
    rocker --mount ~/.vimrc:~/.vimrc ~/.bashrc:~/.bashrc -- ubuntu:18.04
```

src/rocker/mount_extension.py
sloretz: It looks like the default parsing of the docker argument is a lot less convenient. I have a folder `~/ws/rocker`. Without this change I can run `rocker --mount ~/ws/rocker ubuntu:18.04 bash` and see the content in `~/ws/rocker`.

With this this change, the same command creates an empty folder `/home/sloretz/ws/rocker` inside the container. To get the same behavior as before I have to run `rocker --mount ~/ws/rocker:/home/sloretz/ws/rocker ubuntu:18.04 bash`.

README.md
sloretz: If volume args are passed unmodified then this becomes `~/.vimrc:/home/username/.vimrc ...` because `~` appears to only get expanded at the start of an argument.

README.md
sloretz: 209811986ce5f0a43ea9c8017eb94fd8f03b75e4

src/rocker/mount_extension.py
meyerj: I suggested a solution in https://github.com/sloretz/rocker/pull/1.

sloretz:
I don't think I'll come back to this one anytime soon, so I'll close this for now. I've been using singularity instead of docker and rocker for a while now.

Issue #34, opened 2019-02-14
detect missing docker permissions and provide a useful prompt


tfoote:
Right now you get a long traceback if you don't have permissions to build the docker container

```
$ rocker docker/hello-world                                                    
Plugins found: ['dev_helpers', 'home', 'nvidia', 'pulse', 'user']
Active extensions []                                                           
Writing dockerfile to /tmp/tmp93bgqbui/Dockerfile
vvvvvv                                                                                                

FROM docker/hello-world

^^^^^^
Building docker file with arguments:  {'path': '/tmp/tmp93bgqbui', 'rm': True, 'decode': True, 'nocache': False, 'tag': 'rocker_docker/hello-world'}
Traceback (most recent call last):
  File "/usr/lib/python3/dist-packages/urllib3/connectionpool.py", line 601, in urlopen
    chunked=chunked)
  File "/usr/lib/python3/dist-packages/urllib3/connectionpool.py", line 357, in _make_request
    conn.request(method, url, **httplib_request_kw)
  File "/usr/lib/python3.6/http/client.py", line 1239, in request
    self._send_request(method, url, body, headers, encode_chunked)
  File "/usr/lib/python3.6/http/client.py", line 1285, in _send_request
    self.endheaders(body, encode_chunked=encode_chunked)
  File "/usr/lib/python3.6/http/client.py", line 1234, in endheaders
    self._send_output(message_body, encode_chunked=encode_chunked)
  File "/usr/lib/python3.6/http/client.py", line 1026, in _send_output
    self.send(msg)
  File "/usr/lib/python3.6/http/client.py", line 964, in send
    self.connect()
  File "/usr/lib/python3/dist-packages/docker/transport/unixconn.py", line 33, in connect
    sock.connect(self.unix_socket)
PermissionError: [Errno 13] Permission denied

During handling of the above exception, another exception occurred:

Traceback (most recent call last):
  File "/usr/lib/python3/dist-packages/requests/adapters.py", line 440, in send
    timeout=timeout
  File "/usr/lib/python3/dist-packages/urllib3/connectionpool.py", line 639, in urlopen
    _stacktrace=sys.exc_info()[2])
  File "/usr/lib/python3/dist-packages/urllib3/util/retry.py", line 357, in increment
    raise six.reraise(type(error), error, _stacktrace)
  File "/usr/lib/python3/dist-packages/six.py", line 692, in reraise
    raise value.with_traceback(tb)
[...]

PR #47, opened 2019-07-09
Permissions check

tfoote:
This is a fix for #34 It adds the necessary exception checking for any calls that might fail, as well as checking up front that the daemon is reachable before starting work.

PTAL @sloretz

### DIFF
--- a/src/rocker/cli.py
+++ b/src/rocker/cli.py
@@ -16,9 +16,11 @@
 
 import argparse
 import os
+import requests.exceptions
 import sys
 
 from .core import DockerImageGenerator
+from .core import get_docker_client
 from .core import list_plugins
 from .core import pull_image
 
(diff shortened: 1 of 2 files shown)


src/rocker/cli.py
```diff
@@ -37,6 +39,16 @@ def main():
...
     parser.add_argument('--devices', nargs='*')
 
+    # Check for docker access
+    client = get_docker_client()
+    try:
+        client.containers()
+    except requests.exceptions.ConnectionError as ex:
+        print('Aborting: Could not talk to the docker daemon.')
+        print(ex)
+        print('Do you have permissions to run docker such as being in the docker group?')
```
sloretz: Recommend `sys.stderr.write(msg + '/n')`

src/rocker/cli.py
```diff
@@ -37,6 +39,16 @@ def main():
     parser.add_argument('--network', choices=['bridge', 'host', 'overlay', 'none'])
     parser.add_argument('--devices', nargs='*')
 
+    # Check for docker access
+    client = get_docker_client()
+    try:
+        client.containers()
```
sloretz: Recommend moving this check to after `parser.parse_args()` so `rocker --help` works without trying to connect to a docker daemon.

src/rocker/core.py
```diff
@@ -114,7 +115,11 @@ def build(self, **kwargs):
                 else:
                     print("no more output and success not detected")
                     return 2
- 
+            except requests.exceptions.ConnectionError as ex:
+                print("Docker build failed\n", ex)
+                print(ex)
+                print('Do you have permissions to run docker such as being in the docker group?')
```
sloretz: I can't get this check or the one in `pull` to trigger from the CLI. I think the check in `main()` is sufficient.

tfoote:
Closing this for a refactored version in #96

PR #50, opened 2019-08-22
Fix using rocker for non-root containers

meyerj:
If the base image given by the user does not have root privileges, building the rockerized image will fail. Extensions seem to expect that their Dockerfile snippets are executed with root privileges.

This patch injects a `USER root` instruction right after `FROM base_image`. It is a no-op if the base image already runs as the `root` user.

For similar reasons the `user` extension has to run last, because its snippet changes the active user again and subsequent snippets from other extensions might fail. But this is a special case only. Maybe the extension API should have a more general way to declare dependencies of extensions. For example, `nvidia` depends on `x11`. Without specifying both, I cannot run 3D-accelerated GUI applications in the rockerized container.

### DIFF
--- a/src/rocker/cli.py
+++ b/src/rocker/cli.py
@@ -46,7 +46,7 @@ def main():
     args_dict = vars(args)
     
     active_extensions = [e() for e in plugins.values() if args_dict.get(e.get_name())]
-    # Force user to end if present otherwise it will 
+    # Force user to end if present otherwise it will break other extensions
     active_extensions.sort(key=lambda e:e.get_name().startswith('user'))
     print("Active extensions %s" % [e.get_name() for e in active_extensions])
 
--- a/src/rocker/core.py
+++ b/src/rocker/core.py
@@ -180,9 +180,6 @@ def generate_dockerfile(extensions, args_dict, base_image):
         dockerfile_str += el.get_preamble(args_dict) + '\n'
     dockerfile_str += '\nFROM %s\n' % base_image
     dockerfile_str += 'USER root\n'
-    if 'user' in extensions:
-        extensions.remove('user')
-        extensions.append('user')
     for el in extensions:
         dockerfile_str += '# Snippet from extension [%s]\n' % el.name
         dockerfile_str += el.get_snippet(args_dict) + '\n'
--- a/src/rocker/core.py
+++ b/src/rocker/core.py
@@ -179,6 +179,10 @@ def generate_dockerfile(extensions, args_dict, base_image):
         dockerfile_str += '# Preamble from extension [%s]\n' % el.name
         dockerfile_str += el.get_preamble(args_dict) + '\n'
     dockerfile_str += '\nFROM %s\n' % base_image
+    dockerfile_str += 'USER root\n'
+    if 'user' in extensions:
+        extensions.remove('user')
+        extensions.append('user')
     for el in extensions:
         dockerfile_str += '# Snippet from extension [%s]\n' % el.name
         dockerfile_str += el.get_snippet(args_dict) + '\n'
--- a/src/rocker/extensions.py
+++ b/src/rocker/extensions.py
@@ -142,6 +142,6 @@ class User(RockerExtension):
     def register_arguments(parser):
         parser.add_argument(name_to_argument(User.get_name()),
             action='store_true',
-            help="mount the users home directory")
+            help="run the container with the uid/gid of the host user")
 
 
--- a/src/rocker/templates/user_snippet.Dockerfile.em
+++ b/src/rocker/templates/user_snippet.Dockerfile.em
@@ -19,7 +19,7 @@ RUN existing_user_by_uid=`getent passwd "@(uid)" | cut -f1 -d: || true` && \
 
 @[if not home_extension_active ]@
 # Making sure a home directory exists if we haven't mounted the user's home directory explicitly
-RUN mkhomedir_helper @(name)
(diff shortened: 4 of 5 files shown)


meyerj:
> Yeah, the user already has special logic to execute at the end I think that it will always be special.

I removed the redundant reordering of extensions in https://github.com/osrf/rocker/pull/50/commits/9df8a8cf6b20ec677a4cd1d2e9148b3d5b9b2ecb.

We also found that not all special cases were covered (https://github.com/osrf/rocker/pull/50/commits/054cbc444620ca68f63d7aa7a50077c35b503038)
- A user with the same UID already exists in the container image.
- A group with the same GID already exists in the container image.
- A user with the same name already exists in the container image.

In all those cases the existing user or group gets deleted now before creating a new one.

Changing the password and adding the user to the `sudo` group is not necessary with `/etc/sudoers.d/rocker`. Some usernames might not satisfy the password complexity checks done inside the container depending on the container OS and its configuration.

tfoote (approved):
Thanks for extending the support to more of the corner cases. 

Do you have public example images that could be used as test cases for some of these corner cases?

meyerj:
> Thanks for extending the support to more of the corner cases.
> 
> Do you have public example images that could be used as test cases for some of these corner cases?

I do not have one that I can share as-is, and also did not find a public example on [Docker Hub](https://hub.docker.com/). But I created a simple `Dockerfile` based on [ubuntu:latest]() and pushed it as [meyerj/unprivileged](https://hub.docker.com/r/meyerj/unprivileged):

```Dockerfile
FROM ubuntu
RUN useradd docker
USER docker
CMD ["bash"]
```

Without this patch:
```sh
$ rocker --user meyerj/unprivileged
[...]
Building docker file with arguments:  {'path': '/tmp/tmpjoilklgz', 'rm': True, 'nocache': False}
building > Step 1/5 : FROM meyerj/unprivileged
building >  ---> 22d887b3ed36
building > Step 2/5 : RUN if ! command -v sudo >/dev/null; then       apt-get update       && apt-get install -y sudo       && apt-get clean;     fi
building >  ---> Running in c9430b165011
building > Reading package lists...
building > E: List directory /var/lib/apt/lists/partial is missing. - Acquire (13: Permission denied)

no more output and success not detected
Build failed exiting
```

With this patch:
```sh
$ rocker --user meyerj/unprivileged
[...]
building > Step 1/6 : FROM meyerj/unprivileged
building >  ---> 22d887b3ed36
building > Step 2/6 : USER root
building >  ---> Using cache
building >  ---> 5a200b7615ba
building > Step 3/6 : RUN if ! command -v sudo >/dev/null; then       apt-get update       && apt-get install -y sudo       && apt-get clean;     fi
building >  ---> Using cache
building >  ---> 3e286e07cd50
building > Step 4/6 : RUN existing_user_by_uid=`getent passwd "1000006" | cut -f1 -d: || true` &&     if [ -n "${existing_user_by_uid}" ]; then userdel -r "${existing_user_by_uid}"; fi &&     existing_user_by_name=`getent passwd "johannes" | cut -f1 -d: || true` &&     if [ -n "${existing_user_by_name}" ]; then userdel -r "${existing_user_by_name}"; fi &&     existing_group_by_gid=`getent group "1000" | cut -f1 -d: || true` &&     if [ -z "${existing_group_by_gid}" ]; then       groupadd -g "1000" "johannes";     fi &&     useradd --no-log-init --no-create-home --uid "1000006" -s "/bin/bash" -c "Johannes Meyer,,," -g "1000" -d "/home/johannes" "johannes" &&     echo "johannes ALL=NOPASSWD: ALL" >> /etc/sudoers.d/rocker
building >  ---> Using cache
building >  ---> 0e80cb597379
building > Step 5/6 : RUN mkdir -p "$(dirname "/home/johannes")" && mkhomedir_helper johannes
building >  ---> Using cache
building >  ---> 03cfd1c13f33
building > Step 6/6 : USER johannes
building >  ---> Using cache
building >  ---> 17794f1d1625
building > Successfully built 17794f1d1625
Executing command: 
docker run -it   --rm      17794f1d1625 
johannes@3a3b8efd5ac5:/$ 
```

Even without `--user` some extensions expect that their snippets run with root privileges, e.g. `--nvidia`:
```sh
$ rocker --nvidia meyerj/unprivileged
[...]
running,  docker run -it --rm f31d2d29af59
output:  standard_init_linux.go:211: exec user process caused "permission denied"

/tmp/detect_os failed:
> standard_init_linux.go:211: exec user process caused "permission denied"
WARNING unable to detect os for base image 'meyerj/unprivileged', maybe the base image does not exist
```

But I also noticed a downside, namely if `USER root` is not actually required:
```sh
$ rocker meyerj/unprivileged
Building docker file with arguments:  {'path': '/tmp/tmpgppwbqbt', 'rm': True, 'nocache': False, 'pull': False}
building > Step 1/2 : FROM meyerj/unprivileged
building >  ---> 22d887b3ed36
building > Step 2/2 : USER root
building >  ---> Using cache
building >  ---> 5a200b7615ba
building > Successfully built 5a200b7615ba
Executing command: 
docker run -it   --rm      5a200b7615ba 
root@a3fde251aeda:/# 
```

So I guess a potential follow-up could be to extend class `RockerExtension` with a method that lets extensions specify whether they require root privileges or not, such that `USER root` is injected only if it is required by at least one active extension? Or each of those extension could add `USER root` individually in the snippet returned by `get_snippet()`, with the disadvantage that it might be added more than once.

PR #52, opened 2019-08-27
Enable rocker to be used with images pulled from private registries

meyerj:
Rocker currently does not work with custom registries, where the image name has a `hostname[:port]/` prefix. This is because of the way rocker constructs the names of customized images, by simply prepending the string `rocker_` and a suffix listing the active extension.

But a docker image reference has to match a [specific pattern](https://github.com/docker/distribution/blob/master/reference/reference.go). The names generated by rocker violated this syntax and caused errors for image names like `my.private.registry:5000/user/image:version1`, because `rocker_my.private.registry:5000/user/image:version1` is not a valid reference.

This pull request proposes a new way to tag the image names generated by rocker. Only the tag part is modified, while all other parts of the image reference are not modified. Because tags can only contain alphanumeric characters, dots and dashes, the separation character has been changed from `_` to `-`. So the "rockerized" image name of `tfoote/drone_demo` is `tfoote/drone_demo:rocker-latest-nvidia-pulse-x11-user`:

```
REPOSITORY                                        TAG                                   IMAGE ID            CREATED             SIZE
tfoote/drone_demo                                 rocker-latest-nvidia-pulse-x11-user   2be11d950222        4 minutes ago       10.5GB
tfoote/drone_demo                                 rocker--detection-latest              96831aeb6936        5 minutes ago       10.5GB
tfoote/drone_demo                                 latest                                09eb7cfa2fef        7 hours ago         10.5GB
tfoote/drone_demo                                 2019_Apr_17_1219                      9febf26b027d        4 months ago        7.24GB
```

Probably there is an easier way to implement the reference parser, but I have not found an existing Python module and a simple search and split at `:` would not cover all cases.

I ignored the fact that an image name could also have an optional digest part for now (`docker/image@sha256:xxxxxxxxx`).

### DIFF
--- a/src/rocker/core.py
+++ b/src/rocker/core.py
@@ -28,6 +28,8 @@ import em
 import docker
 import pexpect
 
+from .utils import tag_image_name
+
 class RockerExtension(object):
     """The base class for Rocker extension points"""
 
@@ -76,11 +78,10 @@ class DockerImageGenerator(object):
 
         self.dockerfile = generate_dockerfile(active_extensions, self.cliargs, base_image)
         # print(df)
-        self.image_name = "rocker_" + base_image
+        self.image_name = tag_image_name(base_image, prefix="rocker-")
         if self.active_extensions:
-            self.image_name += "_%s" % '_'.join([e.name for e in active_extensions])
+            self.image_name += "-%s" % '-'.join([e.name for e in active_extensions])
 
-    
     def build(self, **kwargs):
         with tempfile.TemporaryDirectory() as td:
             df = os.path.join(td, 'Dockerfile')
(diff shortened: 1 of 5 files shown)


meyerj:
> Thanks this looks like a good improvement.
> 
> Could you rebase it on the recent changes to get the CI fixes and add a couple test cases that will cover at least the tag name types that you're planning to use. Regexs are one of the hardest things to read and make sure that they keep working for future changes.

Done. I am not very familiar with `nose`, just copied from the other examples. The new test case indeed only tests some corner cases when parsing image names, but does not actually involve the docker service.

tfoote:
Replaced by #60

PR #54, opened 2019-08-29
Use exec system call instead of pexpect to fix terminal width in the container

meyerj:
Was there a specific reason to use `pexpect` to interact for `docker run`? That caused issue with the detection of the terminal width (and probably other terminal settings) inside the running container, I assume because `WINCH` signals were not passed through.

Maybe there are other ways to fix that with `pexpect`, but my simple proposed solution is to use the `exec` system call and replace the Python process with `docker run`, because it is the last thing it is doing anyway.

It would be cleaner to collect the `docker run` arguments from extensions in an array instead of a string, that then needs to be decomposed according to shell syntax.

### DIFF
--- a/src/rocker/core.py
+++ b/src/rocker/core.py
@@ -16,6 +16,7 @@
 
 from collections import OrderedDict
 import os
+import shlex
 import sys
 
 import pkg_resources
@@ -26,7 +27,6 @@ import tempfile
 import em
 
 import docker
-import pexpect
 
 class RockerExtension(object):
     """The base class for Rocker extension points"""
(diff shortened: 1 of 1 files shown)


tfoote:
Overall I was hoping to make more use of pexpect's functionality and it was much better integrated than trying to provide the interactive terminal through the subprocess. I hadn't considered the lower level exec family as I'd generally thought that they were deprecated in favor of the subprocess module. 

There's a ticket for terminal width tracking in #7 with potential solution in #57 if that doesn't resolve it this is probably the best approach to make the terminal experience better.

ruffsl:
This seem to work better than my PR https://github.com/osrf/rocker/pull/57, given that launching byobu starts out with the correct size, rather than having to change the size of the xterm window to get byobu to update out of the default `COLUMNS=80`.

tfoote:
I fixed up #57 for the initial conditions and unit tests. If there are more issues we can switch back to the older exec methods, but for now I'll close this in favor of #57.

Issue #7, opened 2019-01-07
pass through window resize logic to interactive window


tfoote:
By default most things go through except for the window resize signals. 

https://pexpect.readthedocs.io/en/stable/api/pexpect.html#pexpect.spawn.interact

ruffsl:
I'd like to help, but I don't know what is going on. Could you elaborate on what you've linked to and what signals are missing. For example, if I set the command directive to byobu of tmux in the Dockerfile:

``` Dockerfile
CMD ["byobu"]
```

The following two command do not behave the same, in that byobu from `docker run` will resize with to current xterm window size as I change it, while the byobu session from `rocker` will default to `COLUMNS=80`, `LINES=23`.

```
$ docker run -it <image>:<tag>
$ rocker <image>:<tag>
```

PR #57, opened 2019-09-04
Add sigwinch passthrough for xterm resizing

ruffsl:
This seem to help pass resize single events from xterm. Fixes: https://github.com/osrf/rocker/issues/7

The em package import is doing some odd stuff by overwriting sys.stdout, so I just stored it to a global variable before `em.expand()` gets called somewhere and changes it to an instance of `ProxyFile()`.

https://github.com/dirk-thomas/empy/blob/0df7f44312071b976d202a3f8a5fd2ce6aacd174/em.py#L2669

When setting byobu as from https://github.com/osrf/rocker/issues/7#issuecomment-527700022 , I still don't know why it still defaults to COLUMNS=80, LINES=23 for `rocker` and not `docker run`.

### DIFF
--- a/src/rocker/core.py
+++ b/src/rocker/core.py
@@ -24,10 +24,16 @@ import pkgutil
 import subprocess
 import tempfile
 
-import em
-
 import docker
 import pexpect
+
+import fcntl
+import signal
+import struct
+import termios
+
+SYS_STDOUT = sys.stdout
+
 
 class RockerExtension(object):
     """The base class for Rocker extension points"""
@@ -176,6 +182,13 @@ class DockerImageGenerator(object):
                 print("Executing command: ")
                 print(cmd)
                 p = pexpect.spawn(cmd)
+                def sigwinch_passthrough (sig, data):
+                    s = struct.pack("HHHH", 0, 0, 0, 0)
+                    a = struct.unpack('hhhh', fcntl.ioctl(SYS_STDOUT.fileno(),
+                        termios.TIOCGWINSZ , s))
+                    if not p.closed:
+                        p.setwinsize(a[0],a[1])
+                signal.signal(signal.SIGWINCH, sigwinch_passthrough)
                 p.interact()
                 p.terminate()
                 return p.exitstatus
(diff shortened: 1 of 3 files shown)


ruffsl:
Could a new release be synced to the apt/pypi repos soon?
I'd like use the recent improvements for the sros2 demo:
https://github.com/ros-swg/turtlebot3_demo

tfoote:
v0.1.7 is out, and propogating to ros and ros2 repos. I was going to try to get #50 in but went ahead with this for now since it has a conflict.

ruffsl:
@tfoote , could you try the demo I linked to above? Even after upgrading to v0.1.7, I'm still not get the terminal retain the original terminal size on startup like that I checked was working for https://github.com/osrf/rocker/pull/54 . 

Additionally, it seems that changing the size of there terminal no longer works to correct the size, unlike my the start of my original PR here:

<details>

## Before Startup
image

## On Startup
image

## On Resize
image

ruffsl:
In comparison when using https://github.com/osrf/rocker/pull/54:

<details>

```
pip3 uninstall --user rocker
pip3 install --user git+https://github.com/meyerj/rocker.git@use-os-exec-to-run-docker-to-fix-terminal-width
```

## On Startup
image

## On Resize
image

meyerj:
This solution also does not work for me and the `COLUMNS` environment variable is fixed to 80.

Could someone explain why pexpect with kind of complex signal pass-through is preferred rather than a simple exec system call (https://github.com/osrf/rocker/pull/54#issuecomment-530605494)? Overall, Rocker's job is done with the launch of `docker run` and there is no need to keep the Python process alive or to even interact with the docker client running as a child process.

tfoote:
@ruffsl  Sorry I screwed up when I reviewed the diff and removed the critical element of saving the raw sys.stdout earlier. It still passes the tests but clearly doesn't capture the actual output. FIx in #66 

@meyerj When used as just the command line the docker run command is the last thing to happen. However I'm also designing this to be used as a library launching demos to that end I need it to come back. And in the future pexpect's functionality can be used for things like supporting validating that the demos have run successfully.

PR #58, opened 2019-09-06
Add new extensions git and ssh

meyerj:
`--git` mounts the Git config from the host into the container.
`--ssh` forwards the SSH agent from the host into the container (for public key authentication).

Both extensions work with or without the `--user` extension, given that the base image runs as user `root`.

Both together allow to work with Git (commit, push and pull) in the same way as on the host, given that the respective packages are installed in the container.

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -43,9 +43,11 @@ kwargs = {
         'rocker.extensions': [
             'dev_helpers = rocker.extensions:DevHelpers',
             'env = rocker.extensions:Environment',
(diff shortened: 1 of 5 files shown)


tfoote:
There was a slowdown for apt installations and when we build the docker images in the tests it looks like they were timing out. I've retriggered all the builds for the pending PRs.

tfoote:
CI is passing now, but it would be great to have at least a few simple unit tests to make sure that we keep the coverage level up would be great for this.

PR #59, opened 2019-09-10
Do not build and run detect_os() twice in the nvidia extension

meyerj:
The detector image was built twice, once for each call of `get_environment_subs()` from `get_preamble()` and `get_snippet()`. I do not see a reason why the base_image argument would change in between these two calls.

### DIFF
--- a/src/rocker/nvidia_extension.py
+++ b/src/rocker/nvidia_extension.py
@@ -92,16 +92,17 @@ class Nvidia(RockerExtension):
             self._env_subs['user_id'] = os.getuid()
             self._env_subs['username'] = getpass.getuser()
         
-            dist, ver, codename = detect_os(cliargs['base_image'])
-            self._env_subs['image_distro_id'] = dist
-            if self._env_subs['image_distro_id'] not in self.supported_distros:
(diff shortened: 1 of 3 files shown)


tfoote:
This looks to be causing a failure of the nvidia unit tests expecting the old behavior.

meyerj:
> The unit tests will need to be resolved.

Done in https://github.com/osrf/rocker/pull/59/commits/5bc6c792ecc46632c6bc07d12cc6159938d7f0c8.

meyerj:
> The unit tests will need to be resolved.

I reverted the original patch in https://github.com/osrf/rocker/pull/59/commits/5ba3b278ee408e32711bbc0f73d443ba4c56f1dc and implemented caching inside `detect_os()` in https://github.com/osrf/rocker/pull/59/commits/ae7bc97ac284037fc96605ace70c014a7adc7e21.

Although it is unlikely that this can happen outside of unit testing the `nvidia_test.py` reuses the same instance of the plugin to build with Ubuntu 16.04 and 18.04 base images sequentially. The first version of the patch which simply cached all environment variables inside the plugin did not work as expected for this case.

PR #60, opened 2019-09-10
Do not tag intermediate images to allow for automatic pruning

meyerj:
Docker images need to be cleaned up from time to time to not fill up the file system with obsolete images. This is typically done by running `docker image prune` or `docker system prune`. This command does not remove tagged images by default, unless called with `-a`. The latter is typically not what you want, because it also removes all the images pulled from Docker Hub or other remote registries.

In fact, there is no need for Rocker to assign tags to the intermediate images. The image IDs work just fine.

This patch also renders https://github.com/osrf/rocker/pull/52 obsolete.

### DIFF
--- a/src/rocker/cli.py
+++ b/src/rocker/cli.py
@@ -22,7 +22,6 @@ from .core import DockerImageGenerator
 from .core import list_plugins
 from .core import pull_image
 
-from .os_detector import build_detector_image
 from .os_detector import detect_os
 
 
@@ -71,9 +70,6 @@ def detect_image_os():
 
     args = parser.parse_args()    
 
-    if not build_detector_image():
-        print("Failed to build detector.")
-        return 1
     results = detect_os(args.image)
     print(results)
     if results:
--- a/src/rocker/core.py
+++ b/src/rocker/core.py
@@ -16,6 +16,7 @@
 
 from collections import OrderedDict
 import os
+import re
 import sys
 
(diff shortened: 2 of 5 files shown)


tfoote:
The CI on this is failing on Python 3.5 https://travis-ci.org/osrf/rocker/jobs/583145753

```
$ xvfb-run nosetests --with-coverage --cover-package rocker --exclude test_nvidia_glmark2
E..EE.EEE......EEE.E.EE
======================================================================
ERROR: test_device (test_core.RockerCoreTest)
----------------------------------------------------------------------
Traceback (most recent call last):
  File "/home/travis/build/osrf/rocker/test/test_core.py", line 92, in test_device
    self.assertEqual(dig.build(), 0)
  File "/home/travis/build/osrf/rocker/src/rocker/core.py", line 122, in build
    output_callback=lambda output: print("building > %s" % output)
  File "/home/travis/build/osrf/rocker/src/rocker/core.py", line 87, in docker_build
    image_id = match[1]
TypeError: '_sre.SRE_Match' object is not subscriptable
-------------------- >> begin captured stdout << ---------------------
Writing dockerfile to /tmp/tmp3y84_107/Dockerfile
```

The Match appears to only work with subscriptable elements in 3.6 and higher: 
https://stackoverflow.com/questions/52404656/sre-sre-match-object-is-not-subscriptable

PR #67, opened 2019-10-08
Work around glibc issue for os detection

tfoote:
Diagnosted from: https://github.com/osrf/car_demo/issues/70

There appears to be an issue with using subprocess when on ros:kinetic that doesn't appear when detecting on ubuntu:xenial due to the different detection code path. This will work around it for now. I think there's something to file upstream with staticx too.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -3,7 +3,6 @@ python:
   - "3.5"
   - "3.6"
 
-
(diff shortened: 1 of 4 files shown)


tfoote:
Manually merged with history cleanup.

PR #68, opened 2019-10-14
Add verbosity outputs

tfoote:
CLI change mostly to check CI.


tfoote:
included in #67 manual merge

PR #71, opened 2019-11-28
Create a home directory if one does not exist for the user

### LINKED ISSUE #42: create empty home directory from skeleton if --user enabled without --home
Otherwise programs that try to use the home directory will get permission denied errors.

Such as osrf/car_demo#56

Resources:

https://askubuntu.com/questions/152707/how-to-make-user-home-folder-after-account-creation
https://askubuntu.com/questions/335961/create-default-home-directory-for-existing-user-in-terminal

It looks like mkhomedir_helper might be available widely enough.
http://man7.org/linux/man-pages/man8/mkhomedir_helper.8.html

tfoote:
This will avoid many errors.
It will not try if the user's home directory is to be mounted by the home extension.
The tool will also not touch any preexisting directory so if it's mounted any other way it should be safe too.

Fixes #42

### DIFF
--- a/src/rocker/extensions.py
+++ b/src/rocker/extensions.py
@@ -135,7 +135,9 @@ class User(RockerExtension):
 
     def get_snippet(self, cliargs):
         snippet = pkgutil.get_data('rocker', 'templates/%s_snippet.Dockerfile.em' % self.name).decode('utf-8')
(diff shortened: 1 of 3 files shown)


ahcorde:
Tested too

PR #77, opened 2020-03-31
Add an Extension Manager class

tfoote:
This avoids boilerplate for alternative CLI parsing and entry points. 

For example see now that extensions can self declare arguments and be 

This builds on top of #76 as with the cleanup of imports it causes those tests to fail otherwise. 

This is still a draft because it needs a few unit tests added for the extension.


tfoote:
Replaced by #82

PR #80, opened 2020-05-01
Add new argument --env-file

meyerj:
Adds support for the `--env-file` argument, in the same way as `docker run`.
See https://docs.docker.com/engine/reference/commandline/run/#set-environment-variables--e---env---env-file.

I had to introduce a new class method `is_active(self, cliargs)` to delegate the check whether an extension is active or not to the plugin code itself. Otherwise it would not be possible that one extension supports multiple command-line arguments.

### DIFF
--- a/src/rocker/cli.py
+++ b/src/rocker/cli.py
@@ -47,8 +47,8 @@ def main():
     args = parser.parse_args()
     args_dict = vars(args)
     
-    active_extensions = [e() for e in plugins.values() if args_dict.get(e.get_name())]
-    # Force user to end if present otherwise it will 
+    active_extensions = [e() for e in plugins.values() if e.is_active(args_dict)]
+    # Force user to end if present otherwise it will break other extensions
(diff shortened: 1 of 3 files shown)


meyerj:
> I had to introduce a new class method `is_active(self, cliargs)` to delegate the check whether an extension is active or not to the plugin code itself. Otherwise it would not be possible that one extension supports multiple command-line arguments.

Just notived that https://github.com/osrf/rocker/pull/77 would also have solved this problem.

tfoote:
It would be great to implement this on top of #77 I've been testing that locally with some custom extensions and haven't had time to loop back to make #77 ready for review. Including an extension blacklist and non-interactive mode flags. I'll try to land those shortly but it shouldn't change the structure if you want to test using that as the base.

tfoote:
I've reimplemented this on top of #77 @meyerj could you try that out?

meyerj:
> I've reimplemented this on top of #77 @meyerj could you try that out?

Yes, #80 works for me with `--env-file`. Thanks!

PR #83, opened 2020-05-08
Add support for focal and buster

tfoote:
We're only using the one files from the images that's arch independent as long as they have libglvnd available it seems to work.

FIxes #78

PR #84, opened 2020-05-08
Enable use of 20.04

emersonknapp:
Tested running on Pop! OS 20.04 was able to launch rviz2 from the osrf/ros2:nightly image

Signed-off-by: Emerson Knapp <emerson.b.knapp@gmail.com>


tfoote:
Thanks this looks like you took the same approach as #83 which was also tested on buster.

emersonknapp:
Oh yeah! Got in a  few minutes before me :) what a fun coincidence. Guess we're all upgrading

PR #96, opened 2020-08-12
Give a good message if docker is not installed or running

tfoote:
This is a simpler replacement for #47

### DIFF
--- a/src/rocker/core.py
+++ b/src/rocker/core.py
@@ -122,7 +122,7 @@ def get_docker_client():
         docker_client.ping()
         return docker_client
     except (docker.errors.APIError, ConnectionError) as ex:
-        raise DependencyMissing('Docker Client failed to connect to docker deamon.'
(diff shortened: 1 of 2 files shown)


src/rocker/core.py
```diff
@@ -107,11 +113,19 @@ def get_active_extensions(self, cli_args):
...
+        try:
+            docker_client = docker.APIClient()
+        except AttributeError:
+            # docker-py pre 2.0
+            docker_client = docker.Client()
+        # Validate that the server is available
+        docker_client.ping()
+        return docker_client
+    except (docker.errors.APIError, ConnectionError) as ex:
+        raise DependencyMissing('Docker Client failed to connect to docker deamon.'
```
ahcorde, line 125:
```suggestion
        raise DependencyMissing('Docker Client failed to connect to docker daemon.'
```

PR #100, opened 2020-10-20
Do not pull all available tags with --pull

meyerj:
`rocker --pull <image>` without a tag pulls all available tags for that image, and hence can fill the disk quickly.

The reason is that the underlying [`docker_client.pull(image,tag=None)`](https://docker-py.readthedocs.io/en/latest/images.html#docker.models.images.ImageCollection.pull) pulls all available tags if no specific tag is specfied:

> If no tag was specified, the method will return a list of Image objects belonging to this repository.

So the removed method `rocker.core.pull_image(image_name)` would have to decompose the URL given at the command line, and either pass the tag separately, or append `:latest` to `image_name` if no other tag was specified.

This patch removes the broken method completely. It is not needed, because the pull flag can simply be forwarded to the [`docker_client.build()`](https://docker-py.readthedocs.io/en/2.5.1/images.html#docker.models.images.ImageCollection.build) command, at least for [docker-py version 2.5.1](https://docker-py.readthedocs.io/en/2.5.1/images.html#docker.models.images.ImageCollection.build) provided by [Ubuntu Bionic](https://packages.ubuntu.com/bionic/python-docker) and [version 1.8.0](https://docker-py.readthedocs.io/en/1.8.0/api/#build) provided by [Ubuntu Xenial](https://packages.ubuntu.com/xenial/python-docker).

Verified with:
```sh
$ docker pull ubuntu:bionic
bionic: Pulling from library/ubuntu
171857c49d0f: Already exists 
419640447d26: Already exists 
61e52f862619: Already exists 
Digest: sha256:646942475da61b4ce9cc5b3fadb42642ea90e5d0de46111458e100ff2c7031e6
Status: Downloaded newer image for ubuntu:bionic
docker.io/library/ubuntu:bionic
$ docker tag ubuntu:bionic ubuntu:latest
$ rocker ubuntu    # expectation: same as ubuntu:latest
[...]
Building docker file with arguments:  {'path': '/tmp/tmppa0oykh2', 'rm': True, 'nocache': False, 'pull': False}
[...]
root@a95d93f22ee6:/# cat /etc/lsb-release 
DISTRIB_ID=Ubuntu
DISTRIB_RELEASE=18.04
DISTRIB_CODENAME=bionic
DISTRIB_DESCRIPTION="Ubuntu 18.04.5 LTS"
root@a95d93f22ee6:/# exit
exit
```
So the version tagged as `ubuntu:latest` is actually `ubuntu:bionic`. `rocker` did not attempt to pull a newer version.

`--pull` still has the desired effect and the docker daemon pulls the image during the build:
```sh
$ rocker --pull ubuntu    # expectation: same as ubuntu:latest
[...]
Building docker file with arguments:  {'path': '/tmp/tmp8zlicrod', 'rm': True, 'nocache': False, 'pull': True}
[...]
root@475275ae66cc:/# cat /etc/lsb-release 
DISTRIB_ID=Ubuntu
DISTRIB_RELEASE=20.04
DISTRIB_CODENAME=focal
DISTRIB_DESCRIPTION="Ubuntu 20.04.1 LTS"
root@475275ae66cc:/# exit
exit
```

Without this patch, the second invocation `rocker --pull ubuntu` would take very long and pull all available tags.

### DIFF
--- a/src/rocker/cli.py
+++ b/src/rocker/cli.py
@@ -20,7 +20,6 @@ import sys
 
 from .core import DockerImageGenerator
 from .core import get_rocker_version
-from .core import pull_image
 from .core import RockerExtensionManager
 
 from .os_detector import detect_os
@@ -57,9 +56,6 @@ def main():
     print("Active extensions %s" % [e.get_name() for e in active_extensions])
 
     base_image = args.image
-
-    if args.pull:
-        pull_image(base_image)
 
     dig = DockerImageGenerator(active_extensions, args_dict, base_image)
     exit_code = dig.build(**vars(args))
--- a/src/rocker/core.py
+++ b/src/rocker/core.py
@@ -197,6 +197,7 @@ class DockerImageGenerator(object):
             arguments['path'] = td
             arguments['rm'] = True
             arguments['nocache'] = kwargs.get('nocache', False)
+            arguments['pull'] = kwargs.get('pull', False)
             print("Building docker file with arguments: ", arguments)
             try:
                 self.image_id = docker_build(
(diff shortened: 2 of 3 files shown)


tfoote (approved):
Thanks it's always great to reduce the amount of code and make it more efficient.

Issue #105, opened 2020-11-19
os detection failing due to paths


tfoote:
Example output `output:  [6] Cannot open self /tmp/staticx-OdKefo/detect_os or archive /tmp/staticx-OdKefo/detect_os.pkg`

There seems to be a recent regression in PyInstaller that's compounded with staticx
* Regression: https://github.com/pyinstaller/pyinstaller/issues/5257  with Fix https://github.com/pyinstaller/pyinstaller/issues/5258 but there's some related traffic on an older ticket: https://github.com/pyinstaller/pyinstaller/issues/2357

I have not found a version of pyinstaller to pin it to that will work. It might be related to needing to pin the bootloader too. https://github.com/pyinstaller/pyinstaller/issues/2357#issuecomment-720347218

This does specifically work with the pyinstaller output but not after being run through staticx

There's a related older ticket here too: https://github.com/JonathonReinhart/staticx/issues/71

---
This fails in my binary installation from debs, but I cannot reproduce it in my workspace

```
$ rocker --nvidia --x11 --user --home --pulse --nocache tfoote/drone_demo 
tfoote@snowman5:~ Last: [130] (0s Seconds)
$ rocker --version
rocker 0.2.2
tfoote@snowman5:~ Last: [0] (0s Seconds)
$ which rocker
/usr/bin/rocker
```

There is an issue that --nocache doesn't pass through to the detector.

tfoote:
This is fixed with the switch to use the go based os detection in #204

PR #106, opened 2020-11-19
Pin pyinstaller to work around regression executing with staticx

tfoote:
Follow up to #105

### DIFF
--- a/src/rocker/nvidia_extension.py
+++ b/src/rocker/nvidia_extension.py
@@ -94,7 +94,7 @@ class Nvidia(RockerExtension):
             self._env_subs['username'] = getpass.getuser()
         
         # non static elements test every time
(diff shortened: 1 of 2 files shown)


adlarkin:
I started seeing the same issue that was described in #105 today. I just tested this branch, and it fixed the issue for me - is there anything else that needs to be fixed, or will this be merged once you get results back from CI? Also, do you know when the next binary package release will come out with this fix?

piraka9011:
Ditto above, is there a timeline for the binary package release?

adlarkin:
> Ditto above, is there a timeline for the binary package release?

I believe it's already out. I just tested a fresh install of `rocker` in a `venv` and don't have this issue anymore. Here's the version I just tested:

```
$ rocker -v
rocker 0.2.3
```

(installed via `pip3 install rocker`)

Issue #109, opened 2020-12-02
rocker does not work with rootless docker


Tuebel:
I have installed [docker rootless](https://docs.docker.com/engine/security/rootless/) on a machine of our computer pool for students. We do not want the students to have root-like access, thus we avoid setting up the docker group.

When trying a simple `rocker --nvidia nvidia/cuda:11.0-base nvidia-smi`, the following error occurs:
```bash
rocker: error: DependencyMissing encountered: Docker Client failed to connect to docker daemon. Please verify that docker is installed and running. As well as that you have permission to access the docker daemon. This is usually by being a member of the docker group.
```
The error occurs when instantiating the `docker.APIClient()` which can not find the rootless docker instance. However, `docker.from_env().api` works and could serve as a drop in replacement without any other code changes. I will open a pull request, when I'm done testing my container.

On the nvidia side, I also had to change the /etc/nvidia-container-runtime/config.toml to allow the device access.
```config
[nvidia-container-cli]
no-cgroups = true
```

tfoote:
Fixed in #110

Could you maybe open a PR to the README to comment for how to support the no-cgroups for rootless. 

Note that I expect to leverage cgroups in the future for things like resource limitations. I don't know what aspects of the nvidia integration you loose turning that off.

tfoote:
Closing this as the issue is fixed in #110

FrankP89:
How is this fixed? 

For me, just running "rocker" still brings up the error.

Running "sudo rocker ..." on the other hand allows the execution.


Is the command meant to be executed as sudo?

PR #110, opened 2020-12-02
docker.from_env() to detect rootless instance

Tuebel:
See #109


tfoote (approved):
Thanks that seems to be a good substitution.

Issue #91, opened 2020-06-07
Default remove container on shutdown


JC-13:
Is it possible to have an option to not docker run with --rm?

tfoote:
An option to add this would be fine. Something like --no-cleanup.
By default we don't want to be generating a large number of images and containers that need to be cleaned up manually.

tfoote:
Resolved in #111

PR #111, opened 2020-12-02
Add --nocleanup option

Tuebel:
Possible fix for #91.
I did not have the time to fully understand the extension architecture, so this fix might be too hacky.
The changes are small but help a lot when using the container for development.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -5,3 +5,4 @@ deb_dist
 *.pyc
 .coverage
 build
+rocker_venv
--- a/src/rocker/cli.py
+++ b/src/rocker/cli.py
@@ -35,6 +35,7 @@ def main():
     parser.add_argument('command', nargs='*', default='')
     parser.add_argument('--noexecute', action='store_true', help='Deprecated')
     parser.add_argument('--nocache', action='store_true')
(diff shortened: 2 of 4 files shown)


tfoote:
Thanks this looks good to add in the core, as the cleanup automation also is in the core. 

Before merging this could you add a quick unit test so make sure we have coverage of the CLI parsing. A dry_run test and asserting that the --rm is present and then not present with and without the option would be great.

Tuebel:
Yes, I will do that probably towards the end of the week.

Tuebel:
I have looked into it and there are probably a few more changes required to make the docker `cmd` testable.
Currently, the `run` function returns the subprocess `returncode` which makes assertions impossible. I see two possibilities:
1. Hijacking the print statements - brittle and no one will understand why a test is failing when a print statement is changed. But most of the existing code would stay untouched.
2. Create a new function `generate_docker_cmd` which returns the final `cmd` string. I would prefer this.

tfoote (approved):
Separating it out to a separate function is much cleaner. Thanks for doing that. This one too appears to conflict soon. I'll try to get to that shortly as well.

chapulina:
Thanks for this! It would be great to get a release that provides this option.

PR #112, opened 2020-12-09
Allow regular expression device strings

peterpolidoro:
It might be nice to allow regular expressions when specifying devices, including all values that match the regular expression. I do not know if such a feature is within the scope of this package, but I have found it to be very useful with multiple devices.

For example, instead of writing this:

`rocker test_image --devices /dev/dri/card0 /dev/ttyACM0 /dev/ttyACM1 /dev/ttyACM2 /dev/ttyACM3 /dev/ttyACM4 /dev/ttyACM5`

You could instead write something like this:

`rocker test_image --devices /dev/dri/card0 /dev/ttyACM[0-5]`

Or even something like this to connect 16 devices:

`rocker test_image --devices /dev/dri/card0 "/dev/ttyACM([0-9]|1[0-5])"`

The code in this pull request seems to work fine, but it is just a suggestion for one way it could be implemented. It uses sre_yield, which adds another dependency unfortunately, but that makes the regular expression matching easy. There may be a much cleaner location for the matching of the regular expression.

I removed the block of code that added devices to the docker_args in core.py. It seemed redundant when using the devices extension. Perhaps that second check is necessary after all if the devices extension is optional, but then it should probably only be performed if the devices extension is not used.

Other arguments, like volumes, may benefit from a regular expression expansion also, perhaps the expansion could happen in a more centralized location so it acts on every argument where it makes sense without needing to be repeated.

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -7,6 +7,7 @@ install_requires = [
     'empy',
     'pexpect',
     'packaging',
+    'sre_yield',
 ]
 
 # docker API used to be in a package called `docker-py` before the 2.0 release
--- a/src/rocker/core.py
+++ b/src/rocker/core.py
@@ -248,14 +248,6 @@ class DockerImageGenerator(object):
 
         docker_args = ''
 
-        devices = kwargs.get('devices', None)
-        if devices:
-            for device in devices:
-                if not os.path.exists(device):
-                    print("ERROR device %s doesn't exist. Skipping" % device)
-                    continue
-                docker_args += ' --device %s ' % device
-
         for e in self.active_extensions:
             docker_args += e.get_docker_args(self.cliargs)
 
--- a/src/rocker/extensions.py
+++ b/src/rocker/extensions.py
@@ -26,6 +26,8 @@ import sys
 
 from .core import get_docker_client
 
+from sre_yield import AllStrings
+
 
 def name_to_argument(name):
     return '--%s' % name.replace('_', '-')
@@ -46,6 +48,8 @@ class Devices(RockerExtension):
     def get_docker_args(self, cliargs):
         args = ''
         devices = cliargs.get('devices', None)
+        if devices:
+            devices = [device for device_re in devices for device in AllStrings(device_re)]
         for device in devices:
             if not os.path.exists(device):
                 print("ERROR device %s doesn't exist. Skipping" % device)
--- a/test/test_extension.py
+++ b/test/test_extension.py
@@ -82,6 +82,13 @@ class DevicesExtensionTest(unittest.TestCase):
         args = p.get_docker_args(mock_cliargs)
         self.assertFalse('--device' in args)
 
+        # Check case for regular expression device
+        mock_cliargs = {'devices': ['/dev/(random|null)']}
+        self.assertEqual(p.get_snippet(mock_cliargs), '')
+        self.assertEqual(p.get_preamble(mock_cliargs), '')
+        args = p.get_docker_args(mock_cliargs)
+        self.assertTrue('--device /dev/random' in args and '--device /dev/null' in args)
+
 
 class HomeExtensionTest(unittest.TestCase):
 


tfoote (changes requested):
The ability to do this makes some sense, however this adds a lot of complexity when you can also leverage that same capability provided by the command line. `rocker --devices /dev/ttyACM[0-5] -- my_image_name` will expand via the bash evaluation. 

I don't have any ACM devices but similarly

```
$ rocker --devices /dev/ttyS* -- ubuntu
Active extensions ['devices']
Writing dockerfile to /tmp/tmpxn_7ngjk/Dockerfile
vvvvvv
# Preamble from extension [devices]


FROM ubuntu
USER root
# Snippet from extension [devices]


^^^^^^
Building docker file with arguments:  {'path': '/tmp/tmpxn_7ngjk', 'rm': True, 'nocache': False, 'pull': False}
building > Step 1/2 : FROM ubuntu
building >  ---> d70eaf7277ea
building > Step 2/2 : USER root
building >  ---> Using cache
building >  ---> 9d6bf5e9f6c9
building > Successfully built 9d6bf5e9f6c9
Executing command: 
docker run -it   --rm    --device /dev/ttyS0  --device /dev/ttyS1  --device /dev/ttyS10  --device /dev/ttyS11  --device /dev/ttyS12  --device /dev/ttyS13  --device /dev/ttyS14  --device /dev/ttyS15  --device /dev/ttyS16  --device /dev/ttyS17  --device /dev/ttyS18  --device /dev/ttyS19  --device /dev/ttyS2  --device /dev/ttyS20  --device /dev/ttyS21  --device /dev/ttyS22  --device /dev/ttyS23  --device /dev/ttyS24  --device /dev/ttyS25  --device /dev/ttyS26  --device /dev/ttyS27  --device /dev/ttyS28  --device /dev/ttyS29  --device /dev/ttyS3  --device /dev/ttyS30  --device /dev/ttyS31  --device /dev/ttyS4  --device /dev/ttyS5  --device /dev/ttyS6  --device /dev/ttyS7  --device /dev/ttyS8  --device /dev/ttyS9  --device /dev/ttyS0  --device /dev/ttyS1  --device /dev/ttyS10  --device /dev/ttyS11  --device /dev/ttyS12  --device /dev/ttyS13  --device /dev/ttyS14  --device /dev/ttyS15  --device /dev/ttyS16  --device /dev/ttyS17  --device /dev/ttyS18  --device /dev/ttyS19  --device /dev/ttyS2  --device /dev/ttyS20  --device /dev/ttyS21  --device /dev/ttyS22  --device /dev/ttyS23  --device /dev/ttyS24  --device /dev/ttyS25  --device /dev/ttyS26  --device /dev/ttyS27  --device /dev/ttyS28  --device /dev/ttyS29  --device /dev/ttyS3  --device /dev/ttyS30  --device /dev/ttyS31  --device /dev/ttyS4  --device /dev/ttyS5  --device /dev/ttyS6  --device /dev/ttyS7  --device /dev/ttyS8  --device /dev/ttyS9    9d6bf5e9f6c9 
root@001f5ddd5a5f:/# exit
```

I can run the range based regexs too:

```
$ rocker --devices /dev/ttyS[1-2]* -- ubuntu
Active extensions ['devices']
Writing dockerfile to /tmp/tmpo9wen04n/Dockerfile
vvvvvv
# Preamble from extension [devices]


FROM ubuntu
USER root
# Snippet from extension [devices]


^^^^^^
Building docker file with arguments:  {'path': '/tmp/tmpo9wen04n', 'rm': True, 'nocache': False, 'pull': False}
building > Step 1/2 : FROM ubuntu
building >  ---> d70eaf7277ea
building > Step 2/2 : USER root
building >  ---> Using cache
building >  ---> 9d6bf5e9f6c9
building > Successfully built 9d6bf5e9f6c9
Executing command: 
docker run -it   --rm    --device /dev/ttyS1  --device /dev/ttyS10  --device /dev/ttyS11  --device /dev/ttyS12  --device /dev/ttyS13  --device /dev/ttyS14  --device /dev/ttyS15  --device /dev/ttyS16  --device /dev/ttyS17  --device /dev/ttyS18  --device /dev/ttyS19  --device /dev/ttyS2  --device /dev/ttyS20  --device /dev/ttyS21  --device /dev/ttyS22  --device /dev/ttyS23  --device /dev/ttyS24  --device /dev/ttyS25  --device /dev/ttyS26  --device /dev/ttyS27  --device /dev/ttyS28  --device /dev/ttyS29  --device /dev/ttyS1  --device /dev/ttyS10  --device /dev/ttyS11  --device /dev/ttyS12  --device /dev/ttyS13  --device /dev/ttyS14  --device /dev/ttyS15  --device /dev/ttyS16  --device /dev/ttyS17  --device /dev/ttyS18  --device /dev/ttyS19  --device /dev/ttyS2  --device /dev/ttyS20  --device /dev/ttyS21  --device /dev/ttyS22  --device /dev/ttyS23  --device /dev/ttyS24  --device /dev/ttyS25  --device /dev/ttyS26  --device /dev/ttyS27  --device /dev/ttyS28  --device /dev/ttyS29    9d6bf5e9f6c9 
root@eec625a3b641:/# exit
```

This has the added benefit of doing things like checking if the files are on the filesystem which the proposed implementation isn't doing.

So I'd suggest not integrating the regex parsing into the module and doing those checks up front. If you're scripting the interface and don't have bash the file walking,  iterating with checks for your required files should be doable at the script level too. Thus not adding more dependencies to this tool. To do this sort of check as people would expect you would want to check for the file presence and give useful feedback if one or more was missing etc.

Does leveraging the regex before execution meet your requirements?

peterpolidoro:
Yes I was thinking of scripting the interface, using udev rules to trigger containers to restart with new device permissions when devices are connected or disconnected from the system. Good point, though, that any extra dependencies should be on my script's side rather than in this tool, since they are not needed when calling from bash. I will close this thanks.

I still think there might be some redundant code in core.py that you might want to remove. Devices are added to docker_args in core.py and then again by the active extension.

I also find it interesting that removing accidentally duplicated code can cause an automated check for percent code coverage to fail. Seems like it might be a little better if the coverage test failed only when non-covered code is added to a file, rather than failing when covered code is removed and that reduces the coverage percentage.

PR #119, opened 2021-02-04
Update nvidia settings documentation - Ubuntu 20.04 support

fevb:
I've tested the nvidia setting with Ubuntu 20.04, kernel 5.4  and nvidia drivers 460 and it works well :+1:

Added a table in the documentation to reflect this, including also the distributions/kernels/drivers which have been tested according to the current documentation.


tfoote (approved):
Thanks confirmed I'm using the 450 on 20.04 too.

PR #126, opened 2021-03-12
add cuda support to images

tfoote:
Currently a draft proof of concept. This will be revised to be more generic. But this should help with visiblity.


woensug-choi:
I've tried with and without nvidia-cuda-dev installation block. Both worked for me. No need to include nvidia-cuda-dev for pre-caching purpose.

tfoote:
Great thanks for the testing. I cleaned that stanza out of the Dockerfile. And I tweaked the name to reflect that it's just Cuda now not cuda-dev. And I tried a fix for the test to get it to pass on presumably non-nvidia hardware in CI.

woensug-choi:
https://github.com/osrf/rocker/pull/126/commits/817461e5a4320c5322339ced51f329fc7de194db includes all that I needed. Now, the image has `nvcc` and all other bin files for the development environment of the Cuda library. Tested with the multibeam sonar https://field-robotics-lab.github.io/dave.doc/contents/dave_sensors/Multibeam-Forward-Looking-Sonar/

woensug-choi:
Something happend??? I don't see `cuda` flag in main branch anymore...

PR #130, opened 2021-03-23
Force env extension to the end, before user

galou:
Signed-off-by: Gaël Écorchard <gael.ecorchard@cvut.cz>

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -39,7 +39,7 @@ kwargs = {
         'console_scripts': [
             'rocker = rocker.cli:main',
             'detect_docker_image_os = rocker.cli:detect_image_os',
-        ],
+	    ],
         'rocker.extensions': [
             'devices = rocker.extensions:Devices',
             'dev_helpers = rocker.extensions:DevHelpers',
@@ -54,7 +54,7 @@ kwargs = {
             'user = rocker.extensions:User',
             'x11 = rocker.nvidia_extension:X11',
         ]
-    },
+	},
     'author': 'Tully Foote',
     'author_email': 'tfoote@osrfoundation.org',
     'keywords': ['Docker'],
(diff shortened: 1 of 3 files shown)


setup.py
```diff
@@ -31,15 +31,15 @@
 
 kwargs = {
     'name': 'rocker',
-    'version': '0.2.3',
+    'version': '0.2.4',
```
tfoote, line 34: Same comments here about changing the version and the unrelated changes.

tfoote (changes requested):
Quick cleanup on the patch, but otherwise this looks good. Moving the logic into the core makes sense. It would be great to add a quick unit test for this logic too to make sure it doesn't break with any future refactoring.

setup.py
galou: Sorry, I don't get it, the other version change was for the minor number but here I changed the patch version. Do you mean that I should let the version to 0.2.3?

galou (commented):
I have to admit that I don't know how to write a unit test to test that the env extension is one before the last when it's supposed to be the sole extension tested.

tfoote:
>  I should let the version to 0.2.3?

Correct. Your patch may be bundled in with other patches and then a release made. The version number will be bumped during the release process and is separate from your changes.

galou:
Reverted the version number and unrelated changes. Should I squeeze commits?

tfoote:
Thanks for the cleanup. 

> I have to admit that I don't know how to write a unit test to test that the env extension is one before the last when it's supposed to be the sole extension tested.

To write the test, you'll need to invoke it with a couple of extensions enabled and check the ordering. 

extending this test https://github.com/osrf/rocker/blob/280b5e13d3a6b85a872c2250dceb489b9ea06dcc/test/test_core.py#L123

Coverage should be decent with the following tests:
* push user to last
* push env to last except user
* push env to second to last with user

PR #132, opened 2021-03-24
Add the privileged extension

galou:
Signed-off-by: Gaël Écorchard <gael.ecorchard@cvut.cz>

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -31,7 +31,7 @@ install_requires.append(docker_package)
 
 kwargs = {
     'name': 'rocker',
-    'version': '0.3.0',
+    'version': '0.2.5',
     'packages': ['rocker'],
     'package_dir': {'': 'src'},
     'package_data': {'rocker': ['templates/*.em']},
@@ -39,7 +39,7 @@ kwargs = {
         'console_scripts': [
             'rocker = rocker.cli:main',
             'detect_docker_image_os = rocker.cli:detect_image_os',
-        ],
+	    ],
         'rocker.extensions': [
             'dev_helpers = rocker.extensions:DevHelpers',
             'devices = rocker.extensions:Devices',
@@ -55,7 +55,7 @@ kwargs = {
             'user = rocker.extensions:User',
             'x11 = rocker.nvidia_extension:X11',
         ]
-    },
+	},
     'author': 'Tully Foote',
     'author_email': 'tfoote@osrfoundation.org',
     'keywords': ['Docker'],
--- a/setup.py
+++ b/setup.py
@@ -31,7 +31,7 @@ install_requires.append(docker_package)
 
 kwargs = {
     'name': 'rocker',
-    'version': '0.2.3',
+    'version': '0.3.0',
     'packages': ['rocker'],
     'package_dir': {'': 'src'},
     'package_data': {'rocker': ['templates/*.em']},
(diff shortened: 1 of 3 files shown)


setup.py
```diff
@@ -31,30 +31,31 @@
 
 kwargs = {
     'name': 'rocker',
-    'version': '0.2.3',
+    'version': '0.3.0',
```
tfoote, line 34: please don't change the version number in a patch. This will be done at release time.

setup.py
```diff
@@ -31,30 +31,31 @@
...
     'packages': ['rocker'],
     'package_dir': {'': 'src'},
     'package_data': {'rocker': ['templates/*.em']},
     'entry_points': {
         'console_scripts': [
             'rocker = rocker.cli:main',
             'detect_docker_image_os = rocker.cli:detect_image_os',
-	    ],
+        ],
```
tfoote, line 42: please avoid unrelated changes to the PR.

src/rocker/extensions.py
```diff
@@ -273,3 +273,26 @@ def register_arguments(parser, defaults={}):
...
+        return 'privileged'
+
+    def __init__(self):
+        self.name = Privileged.get_name()
+
+    def get_snippet(self, cli_args):
+        return ''
+
+    def get_docker_args(self, cli_args):
+        return '--privileged'
```
tfoote, line 291: This needs to be prepended with whitespace to make sure not to run into a preceding argument if not first. 
```suggestion
        return ' --privileged'
```

They are currently string appended https://github.com/osrf/rocker/blob/4f9f6d7595ccc23a2ae9b600a2262deefb676d89/src/rocker/core.py#L257 This could be made more robust with a ' '.join(...) it looks like other places have this potential problem depending on ordering.

setup.py
tfoote: A little cleanup can be done, but it should be at least separated to a different commit.

tfoote (changes requested):
Thanks this will be useful to add. I had a few requests to clean up the patch but otherwise it looks good.

setup.py
galou: My point of view on this is that we add a functionality and thus must increment the minor version number but will obey to your request. I'll increment first RP #130.

src/rocker/extensions.py
galou: Wouldn't it be better to actually let extensions return a list of strings and let core.py manage formatting the command, if needed?

setup.py
tfoote: Yes, that's the policy for semantic versioning. But this commit is not the release. There may be multiple additions that go into the next release. And I couldn't merge this PR at the same time as your other one because this change would conflict.

src/rocker/extensions.py
tfoote: I think a single string at least with core.py inserting the space. "If needed" is not clearly resolved if I return multiple.

setup.py
galou: I get it with the version number. I actually never contributed to a release-based repository. How do you then manage the version change? Do contributors have to notify somewhere that they did a change that provokes a version change?

src/rocker/extensions.py
galou: There might be a few words missing in these last sentences and I don't get their sense. Anyway, with "if needed", I meant that `subprocess.run` and partially `pexcpect.spawn` accept a list of strings as argument to launch the command.

galou:
Reverted the version number and unrelated changes. Should I squeeze commits?

tfoote (approved):
Thanks for cleaning it up. I added some tests too. I'll just wait for CI to return. Don't worry about cleaning up the commit history, I'll squash when merging.

tfoote:
Coverage isn't reporting directly: 100% diff coverage now https://app.codecov.io/gh/osrf/rocker/compare/132/overview

PR #133, opened 2021-03-26
Remove intermediate containers and name detect_os images

stonier:
`detect_os` creates a flurry of images and containers and dumps them on your system, with little introspectability as to what they are. This does a partial cleanup:

* Removes intermediate containers spun up by the dockerfile
* Names the final image (though not the preliminary build image) for the os detector


stonier:
A little baffled re the error - tests are working for me locally. Any ideas?

src/rocker/os_detector.py
```diff
@@ -53,6 +53,13 @@ def detect_os(image_name, output_callback=None, nocache=False):
 
     iof = StringIO((DETECTION_TEMPLATE % locals()).encode())
     image_id = docker_build(fileobj = iof, output_callback=output_callback, nocache=nocache)
+    image_id = docker_build(
```
tfoote, line 56: The double invocation of `docker_build` looks like an oversight?

tfoote (changes requested):
Cleaning intermediate containers sounds good as well as tagging the final one.

src/rocker/os_detector.py
stonier: Oh indeed. Thanks.

stonier: Removed.

tfoote (approved):

meyerj:
Since the temporary images are tagged they would not be removed automatically by [`docker image prune`](https://docs.docker.com/engine/reference/commandline/image_prune/) or [`docker system prune`](https://docs.docker.com/engine/reference/commandline/system_prune/) without the `--all`/`-a` option anymore. Instead, they slowly fill up the disk and require manual cleaning from time to time. Adding `--all` is not an option either, because then all tagged images would be removed and need to be rebuilt or pulled from a remote registry again.

Actually the intention of https://github.com/osrf/rocker/pull/60 was more or less the opposite, i.e. to avoid having tags assigned to temporary, cached images, such that they are considered as "dangling" by docker's builtin cleanup tools.

PR #138, opened 2021-03-30
git extension supporting user override and full coverage

tfoote:
Support for use with #137


tfoote:
Coverage up: https://app.codecov.io/gh/osrf/rocker/compare/138

PR #142, opened 2021-04-17
Addressing rocker not capable of mounting local folders

130s:
Re-opening https://github.com/osrf/rocker/pull/46 and https://github.com/sloretz/rocker/pull/1 as a new PR rebased to the `origin/main` of the moment.

# Issue aimed
There's no (easy) way to mount the host's volume on to the container `rocker` starts.

# Approach
Add a volume option similar to `docker run --volume` option.

# Review checlist includes
- [x] Review comments in https://github.com/osrf/rocker/pull/46 addressed.
- [x] Unit tests added.
- [x] dev-tested.


130s:
I think the review comments in https://github.com/osrf/rocker/pull/46 are all addressed (mostly thanks to @meyerj's https://github.com/sloretz/rocker/pull/1).

README.md
```diff
@@ -139,3 +139,17 @@ On Xenial
...
+
+`--volume` adds paths as docker volumes. The last path must be terminated with two dashes `--`.
+
+    rocker --volume ~/.vimrc ~/.bashrc -- ubuntu:18.04
+
+Same as `docker`'s `--volume` option, `rocker --volume` takes 3 fields.
+- 1st field: the path to the file or directory on the host machine.
+- 2nd field: (optional) the path where the file or directory is mounted in the container.
+   - If only the 1st field is supplied, same value as the 1st field will be populated as the 2nd field.
+- 3rd field: (optional) and is a comma-separated list of options, such as ro, z, and Z. These options are discussed below.
```
tfoote, line 153: This options aren't discussed below. Maybe a link to the docker documentation itself would make sense?

Also a comment on how we're expanding the arguments before passing them through, because that's a slight difference from the default docker approach.

tfoote (commented):
Thanks for pulling all the different elements together.

README.md
130s: @tfoote Thanks, hope I addressed your comment by now.

README.md
```diff
@@ -139,3 +139,19 @@ On Xenial
...
+
+## Volume mount
+
+`--volume` adds paths as docker volumes. The last path must be terminated with two dashes `--`.
+
+    rocker --volume ~/.vimrc ~/.bashrc -- ubuntu:18.04
+
+The above example of the volume option will be expanded in `docker run` as
+
+    --volume ~/.vimrc:~/.vimrc --volume ~/.bashrc:~/.bashrc
```
tfoote, line 151: Thanks for the iteration, but the other important part of this expansion is that it will resolve `~/` to `/home/<USERNAME>/` via abspath. Docker will abort if you try to pass non-absolute paths.

meyerj:
> I think the review comments in #46 are all addressed (mostly thanks to @meyerj's [sloretz#1](https://github.com/sloretz/rocker/pull/1)).

We at Intermodalics are using this extension in the version from [sloretz#1](https://github.com/sloretz/rocker/pull/1) now for a while. Note that there is an ambiguity issue with argparse if there is no other option anymore after the list of folders to mount and the image name. But from the [README](https://github.com/osrf/rocker/pull/142/files#diff-b335630551682c19a781afebcf4d07bf978fb1f8ac04c6bf87428ed5106870f5) it seems this has been addressed here, or argparse can already understand the double-dash to separate the positional arguments.

In the meantime I also added a [patch on an internal version](https://github.com/meyerj/rocker/tree/feature/argument-pass-through) to just pass any argument through to `docker run`, motivated by [x11docker](https://github.com/mviereck/x11docker#terminal-usage):
```
To run a Docker container on a new X server:
  x11docker IMAGE
  x11docker [OPTIONS] IMAGE [COMMAND]
  x11docker [OPTIONS] -- IMAGE [COMMAND [ARG1 ARG2 ...]]
  x11docker [OPTIONS] -- DOCKER_RUN_OPTIONS -- IMAGE [COMMAND [ARG1 ARG2 ...]]
```
That removes the need for some extensions that essentially just mimic the command-line usage of `docker run`, like this one or the `Environment` extension, and allows to apply special options like for memory or CPU limits, custom container IPs, start detached etc. I still wanted to submit a PR for that, if you are interested. For `--mount`/`--volume` the possibility to pass a single folder and mount it at the same path inside the container is convenient and not supported like that by `docker run`.

README.md
tfoote, line 151:
```suggestion
    --volume /home/<USERNAME>/.vimrc:/home/<USERNAME>/.vimrc --volume /home/<USERNAME>/.bashrc:/home/<USERNAME>/.bashrc  
```

README.md
```diff
@@ -139,3 +139,19 @@ On Xenial
...
 
     rocker --nvidia --x11 osrf/ros:melodic-desktop-full gazebo
+
+## Volume mount
+
+`--volume` adds paths as docker volumes. The last path must be terminated with two dashes `--`.
+
+    rocker --volume ~/.vimrc ~/.bashrc -- ubuntu:18.04
+
+The above example of the volume option will be expanded in `docker run` as
```
tfoote, line 149:
```suggestion
Using the shorter version in the above example of the volume option will be expanded using the absolute path before being passed to `docker run` like this:
```

tfoote:
Squashed and manually merged in dc046c7

PR #149, opened 2021-05-25
Fix os detection for non-root images

miguelprada:
The `detect_os` executable has execution permission set for it's owner (i.e. `root`) only. When attempting to run an image that has a default non-`root` user, this causes a _permission denied_ error in the [entrypoint](https://github.com/osrf/rocker/blob/9d7a3924cd0d9a4c45ea163a7ab5296e4cdfdd3c/src/rocker/os_detector.py#L43).

This simple fix should solve the issue.


tfoote (changes requested):
Thanks this makes sense. I'd like to collapse it into one line if possible though not to incur another docker layer.

Can you also provide a quick reproduction example to test it with and without the change?

miguelprada:
Continues on #150.

PR #150, opened 2021-06-07
Fix os detection for non-root images (cont.)

miguelprada:
This is a continuation of #149. I made the mistake of opening the PR directly from a branch in my colleague's fork without realizing I don't have push access to that. Apologies for the inconvenience.

I think this PR also addresses the [comment](https://github.com/osrf/rocker/pull/149#pullrequestreview-676744607) by @tfoote by appending the `chmod` onto the previous `RUN` command.

As for the reproduction example, you can trigger the error by creating an image with a non-root user, and then try to run that image with the `--nvidia` flag.

For instance, this

```bash
mkdir -p /tmp/context && cd /tmp/context
cat <<EOF > Dockerfile
FROM ros:melodic
RUN useradd -ms /bin/bash someuser
USER someuser
EOF
docker build -t deletethis .
rocker --nvidia deletethis
```

fails for me on `main`, whereas it passes ok with the fix.

### DIFF
--- a/src/rocker/os_detector.py
+++ b/src/rocker/os_detector.py
@@ -37,6 +37,8 @@ RUN . /tmp/distrovenv/bin/activate && pyinstaller --onefile /tmp/distrovenv/dete
 
 RUN . /tmp/distrovenv/bin/activate && staticx /dist/detect_os /dist/detect_os_static
 
+RUN chmod go+xr /dist/detect_os_static
+
 FROM %(image_name)s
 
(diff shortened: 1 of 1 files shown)


tfoote (approved):
Thanks for the iteration. This looks good and I validated it locally.

It's failing CI due to an upstream bug: https://github.com/docker/docker-py/issues/2807  I've got a workaround coming for that in #151

PR #156, opened 2021-08-03
Fix detector for the deprecated distro syntax

### LINKED ISSUE #155: Nvidia extension stopped working
Running a simple rocker command with nvidia (`rocker --nvidia ubuntu`) yields the following error:

```    
detect_os.py:1: DeprecationWarning: distro.linux_distribution() is deprecated. 
It should only be used as a compatibility shim with Python's platform.linux_distribution().
Please use distro.id(), distro.version() and distro.name() instead. 
```

tfoote:
Fixes #155


InigoMoreno:
Thanks!

PR #158, opened 2021-08-13
cuda-dev

woensug-choi:
The rocker cuda PR does not include executables of the Cuda libraries. It's because PR only copied apt-get installation of the Nvidia cudagl's base dockerfile. For our purposes, all those executables are required to compile the .cu source code of the multibeam sonar plugin.

I believe that --cuda flag without Cuda executables could have usability that does not require compile of .cu source files. I've made a new --cuda-dev flag that includes all installations required.

I've tested the functionality with the Dave project's Multibeam Sonar Plugin.
https://github.com/Field-Robotics-Lab/DAVE


woensug-choi:
@tfoote I hope to see this merged or receive comments if you think its not ready :)

src/rocker/templates/cuda_dev_snippet.Dockerfile.em
```diff
@@ -0,0 +1,63 @@
...
+## Required for nvidia-docker v1
+#RUN echo "/usr/local/nvidia/lib" >> /etc/ld.so.conf.d/nvidia.conf \
+#    && echo "/usr/local/nvidia/lib64" >> /etc/ld.so.conf.d/nvidia.conf
+
+ENV PATH /usr/local/nvidia/bin:/usr/local/cuda/bin:${PATH}
+ENV LD_LIBRARY_PATH /usr/local/nvidia/lib:/usr/local/nvidia/lib64${LD_LIBRARY_PATH}
+
+# nvidia-container-runtime
+ENV NVIDIA_VISIBLE_DEVICES all
+ENV NVIDIA_DRIVER_CAPABILITIES compute,utility
```
tfoote, line 58: These options here need to mutate/extend the driver capabilitiy not just overwrite it. aka it could be used for graphics too.

src/rocker/templates/cuda_dev_snippet.Dockerfile.em
```diff
@@ -0,0 +1,63 @@
...
+    gnupg2 curl ca-certificates && \
+    curl -fsSL https://developer.download.nvidia.com/compute/cuda/repos/ubuntu2004/x86_64/7fa2af80.pub | apt-key add - && \
+    echo "deb https://developer.download.nvidia.com/compute/cuda/repos/ubuntu2004/x86_64 /" > /etc/apt/sources.list.d/cuda.list && \
+    echo "deb https://developer.download.nvidia.com/compute/machine-learning/repos/ubuntu2004/x86_64 /" > /etc/apt/sources.list.d/nvidia-ml.list && \
+    apt-get purge --autoremove -y curl \
+    && rm -rf /var/lib/apt/lists/*
+
+ENV CUDA_VERSION 11.2.1
+
+ENV NV_CUDA_LIB_VERSION "11.2.1-1"
```
tfoote, line 17: Having all of these versions embedded seems very fragile. Is there a way to do this more generically? What's the upgrade path, how can we let people adjust the version looking forward to future releases and other platforms?

tfoote (commented):
Overall this looks like a good direction. 

For this to go in it definitely needs some unit tests. And I'd like to have a better story about how the different layers of base, runtime and development can be used together. Aka get this working with #126 The lack of tests and validation of cooperation is why I haven't pushed #126 further forward. 

And also it should not collide/conflict with the generic --nvidia option.

src/rocker/templates/cuda_dev_snippet.Dockerfile.em
woensug-choi: Changed to 
```em
ENV NVIDIA_DRIVER_CAPABILITIES ${NVIDIA_DRIVER_CAPABILITIES:+$NVIDIA_DRIVER_CAPABILITIES,}compute,utility
```

src/rocker/templates/cuda_dev_snippet.Dockerfile.em
woensug-choi: The latest version gets updated constantly. It's now at 11.6.0 (https://gitlab.com/nvidia/container-images/cuda/blob/master/dist/11.6.0/ubuntu2004/base/Dockerfile)

How could we exploit https://gitlab.com/nvidia/container-images/cuda/-/blob/master/manifests/cuda.yaml ? If we could load this during the build and have user to type wanted Cuda version (e.g. 11.6.0), it could be much more general.

woensug-choi:
What kinds of unit tests do we need? I've been using this branch with a gazebo plugin that requires Cuda compilation.

What do you mean by layer story? Both [`devel`](https://gitlab.com/nvidia/container-images/cuda/-/blob/master/dist/11.6.0/ubuntu2004/devel/Dockerfile) and [`runtime`](https://gitlab.com/nvidia/container-images/cuda/-/blob/master/dist/11.6.0/ubuntu2004/runtime/Dockerfile) image based on the [`base`](https://gitlab.com/nvidia/container-images/cuda/blob/master/dist/11.6.0/ubuntu2004/base/Dockerfile).

src/rocker/templates/cuda_dev_snippet.Dockerfile.em
tfoote: As you mention the latest gets updated frequently. How can we use the latest version without hard coding all these many versions in our codebase? if they're changing the version constantly we have to check it works and update our codebase too? How do we get notices and know what to update when? Will people's usage break w/o updates? Will we break users if we update too much? How do our users know/declare what versions they get/want?

tfoote:
For tests, we need tests that provide coverage of the code paths. They need to run examples that exercise the code paths and exercise the cuda capabilities so that we know that it's still working when we deploy it. We have 84% code coverage on the non-nvidia stuff. If you run it locally with the nvidia tests it's higher. But you content has zero unit tests.  

For the story we need to be able to explain to the users. If you want this features do this. If you want that feature do that. If you want this and that t do they work together? In particular since this is taking the capabilities of #126 and going further, it should probably be done as a series of layers that will first enable the base, and then install the specific tools, so that you have core nvidia, then extend it for the runtime, and then extend that for dev tools. Instead of having each plugin potentially installing different separate versions of cuda tools.

tfoote:
There's an overview of how to run the tests and you'll see reports here: https://github.com/osrf/rocker#testing

src/rocker/templates/cuda_dev_snippet.Dockerfile.em
```diff
@@ -0,0 +1,23 @@
...
+    gnupg2 curl ca-certificates && \
+    apt-key del 7fa2af80 && \
+    curl -L -O https://developer.download.nvidia.com/compute/cuda/repos/ubuntu2004/x86_64/cuda-keyring_1.0-1_all.deb && \
+    dpkg -i cuda-keyring_1.0-1_all.deb && \
+    sudo sed -i '/developer\.download\.nvidia\.com\/compute\/cuda\/repos/d' /etc/apt/sources.list && \
+    apt-get purge --autoremove -y curl \
+    && rm -rf /var/lib/apt/lists/*
+
+RUN apt-get update && apt-get install -y --no-install-recommends \
+    cuda nvidia-cuda-dev \
```
woensug-choi, line 11: I've made changes to use the package manager to install necessary packages without defining specific versions. I've tested it inside WSL. I have not tested with a generic Ubuntu system but it should work.

woensug-choi:
https://github.com/osrf/rocker/blob/0dea4de4014cf288e7f914e21b00edef498b1545/src/rocker/templates/cuda_dev_snippet.Dockerfile.em#L10-L12

@tfoote I've made a change to use the package manager. It now avoids specifying versions for dependent packages. I've tested in WSL Ubuntu 20.04. It should also work for the generic Ubuntu installation.

tfoote (commented):
I've spent a few hours to converge #182 and this into the branch for #126

And following the installation instructions from https://developer.nvidia.com/cuda-downloads?target_os=Linux&target_arch=x86_64&Distribution=Ubuntu&target_version=22.04&target_type=deb_network as closely as possible. 

There seems to be an issue with installing cuda and nvidia-cuda-dev simultaneously. But I've run out of time to iterate on that right now so I've pushed it into the draft.

src/rocker/templates/cuda_dev_snippet.Dockerfile.em
```diff
@@ -0,0 +1,23 @@
...
+# File conflict problem with libnvidia-ml.so.1 and libcuda.so.1
+# https://github.com/NVIDIA/nvidia-docker/issues/1551
+RUN rm -rf /usr/lib/x86_64-linux-gnu/libnv*
+RUN rm -rf /usr/lib/x86_64-linux-gnu/libcuda*
+
+ENV PATH /usr/local/cuda/bin${PATH:+:${PATH}}
+ENV LD_LIBRARY_PATH /usr/local/cuda/lib64/stubs:/usr/local/cuda/lib64${LD_LIBRARY_PATH:+:${LD_LIBRARY_PATH}}
+
+ENV NVIDIA_VISIBLE_DEVICES ${NVIDIA_VISIBLE_DEVICES:-all}
+ENV NVIDIA_DRIVER_CAPABILITIES ${NVIDIA_DRIVER_CAPABILITIES:+$NVIDIA_DRIVER_CAPABILITIES,}compute,utility
```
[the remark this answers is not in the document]
woensug-choi: confirmed

src/rocker/templates/cuda_dev_snippet.Dockerfile.em
```diff
@@ -0,0 +1,23 @@
+RUN apt-get update && apt-get install -y --no-install-recommends \
```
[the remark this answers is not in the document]
woensug-choi: I've tried with and without nvidia-cuda-dev installation block. Both worked for me. No need to include nvidia-cuda-dev for pre-caching purpose.

tfoote:
@woensug-choi I think that I've merged all of this into #126 could you verify that it meets your needs now? 

Note: I changed the argument to `--cuda`

woensug-choi:
@tfoote I confirm that everything I need is merged into #126. Closing this PR.

tfoote:
Great thanks for verifying.

PR #159, opened 2021-09-07
Add an option to tag the image

tfoote:
This can make the resultant image easier to use and potentially push to a registry.

### DIFF
--- a/src/rocker/core.py
+++ b/src/rocker/core.py
@@ -103,6 +103,8 @@ class RockerExtensionManager:
         parser.add_argument('--mode', choices=OPERATION_MODES,
             default=OPERATIONS_INTERACTIVE,
             help="Choose mode of operation for rocker")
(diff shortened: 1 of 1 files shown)


src/rocker/core.py
```diff
@@ -223,6 +225,10 @@ def build(self, **kwargs):
             arguments['rm'] = True
             arguments['nocache'] = kwargs.get('nocache', False)
             arguments['pull'] = kwargs.get('pull', False)
+            image_name = kwargs.get('image_name', None)
+            if image_name:
+                print(f"Running docker tag {self.image_id} {image_name}")
```
meyerj, line 230: This patch broke compatibility with Python 3.5 and Ubuntu 16.04, which do not support f-strings yet. Not sure whether you still consider that relevant because it is already end-of-life, but it would be simple to fix by using another string formatting method here?

PR #162, opened 2021-10-05
pin staticx

### LINKED ISSUE #161: nvidia extension: OS detection is failing
I cleared out some cached docker images today that were laying around for a few weeks. As a result, when I try to run the `--nvidia` extension, I get a `detect_os` failure that I wasn't seeing before.

For example, if I run the following command, I have no issues:
```
rocker ubuntu:bionic bash
```

Add the `--nvidia` argument (`rocker --nvidia ubuntu:bionic bash`), and I see the following error:
```
Step 12/12 : CMD [ "" ]
 ---> Running in 165a674db9fa
Removing intermediate container 165a674db9fa
 ---> fbf38e2b1399
Successfully built fbf38e2b1399
Successfully tagged rocker:os_detect_ubuntu_bionic
running,  docker run -it --rm fbf38e2b1399
output:  
/tmp/detect_os failed:
WARNING unable to detect os for base image 'ubuntu:bionic', maybe the base image does not exist
```

I am using rocker version `0.2.4`.

It looks like the error is here: https://github.com/osrf/rocker/blob/main/src/rocker/nvidia_extension.py#L100

Perhaps something has changed in an upstream dependency?

tfoote:
Fixes #161

### DIFF
--- a/src/rocker/os_detector.py
+++ b/src/rocker/os_detector.py
@@ -32,7 +32,7 @@ RUN python3 -m venv /tmp/distrovenv
 # patchelf needed for staticx
 # binutils provides objdump needed by pyinstaller
 RUN apt-get update && apt-get install -qy patchelf binutils
-RUN . /tmp/distrovenv/bin/activate && pip install distro pyinstaller==4.0 staticx
+RUN . /tmp/distrovenv/bin/activate && pip install distro pyinstaller==4.0 staticx==0.12.3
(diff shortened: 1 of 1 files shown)


adlarkin (approved):
This fixed the OS detector  issue for me with both `ubuntu:bionic` and `ubuntu:focal` images. Thanks!

PR #164, opened 2021-11-30
add an option to preserve the home directory instead of deleting it if the uid or username collide

tfoote:
This allows persistence of existing content in home directories from underlays if not using the --home option.


chapulina (approved):
Looks reasonable to me

PR #166, opened 2021-12-13
Volume argument parsing change proposal

darkgap:
This is a proposal, as the the [README.md](https://stackoverflow.com/a/36338204) actually specifies that the last path should be terminated with `--`, but I missed that. This made me lose quite a bit of time understanding what was going on since I was trying to bind volumes as follows:

```
rocker --nvidia --x11 --name iaslab --volume "$HOME/my/folder:/container/folder" user/image
usage: rocker [-h] [--noexecute] [--nocache] [--nocleanup] [--pull] [-v] [--dev-helpers]
              [--devices [DEVICES [DEVICES ...]]] [--env NAME[=VALUE] [NAME[=VALUE] ...]]
              [--env-file ENV_FILE] [--git] [--git-config-path GIT_CONFIG_PATH] [--home]
              [--name NAME] [--network {host,none,bridge}] [--nvidia] [--privileged]
              [--pulse] [--ssh] [--user] [--user-override-name USER_OVERRIDE_NAME]
              [--user-preserve-home] [--volume HOST-DIR[:CONTAINER-DIR[:OPTIONS]]
              [HOST-DIR[:CONTAINER-DIR[:OPTIONS]] ...]] [--x11]
              [--mode {interactive,non-interactive,dry-run}] [--image-name IMAGE_NAME]
              [--extension-blacklist [EXTENSION_BLACKLIST [EXTENSION_BLACKLIST ...]]]
              image [command [command ...]]
rocker: error: the following arguments are required: image
```

The argument parser for the `--volume` flag, being configured with `nargs='+'`, was causing the `--volume` directive to eat up the *image* positional argument if used as last optional one (also see [this StackOverflow issue](https://stackoverflow.com/a/36338204)). 
Adopting this change makes the configuration of volumes so that each bind is specified with its own `--volume` flag.

I think that this approach is slightly better (and also more inline with the way the Docker CLI handles it). If changing the code is not possible or wanted, I propose to document the need for `--` in the `rocker --help` message.

Co-authored with @gdelazzari

### DIFF
--- a/README.md
+++ b/README.md
@@ -142,17 +142,19 @@ On Bionic
 
 ## Volume mount
 
-For arguments with one element not colon separated.
+### For arguments with one element not colon separated
 
-`--volume` adds paths as docker volumes. The last path must be terminated with two dashes `--`.
+`--volume` adds paths as docker volumes. One `--volume` directive must be used for each path.
 
-    rocker --volume ~/.vimrc ~/.bashrc -- ubuntu:18.04
+    rocker --volume ~/.vimrc --volume ~/.bashrc ubuntu:18.04
 
 The above example of the volume option will be expanded via absolute paths for `docker run` as follows:
 
     --volume /home/<USERNAME>/.vimrc:/home/<USERNAME>/.vimrc --volume /home/<USERNAME>/.bashrc:/home/<USERNAME>/.bashrc  
 
-For arguments with colon separation it will process the same as `docker`'s `--volume` option, `rocker --volume` takes 3 fields.
+### For arguments with colon separation
+
+it will process the same as `docker`'s `--volume` option, `rocker --volume` takes 3 fields.
 - 1st field: the path to the file or directory on the host machine.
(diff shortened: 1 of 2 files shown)


tfoote:
Requiring repeating the `--volume` argument makes the command much more verbose. And the use of `--` is a standard of the argparse tool. This is something that's standard upstream. If you're using any `nargs=+` or `nargs=?' arguments this can be an ambiguity. If you use the `--volume` argument before any other argument that starts with a `--` this will not be necessary. 

I believe that most of the argument parsing is automated. But if you can find a way to detect that there's a nargs argument at the end without any positional arguments then it likely is something that would be helpful to the user and potentially even submitted upstream for standard processing. In general I don't want to start doing custom processing of command line arguments.

tfoote (commented):
Moving my comment over here to the review. The structure of the argument parsing is a core upstream capability that I don't expect to change. Improved errors or other features would be good to add.

PR #171, opened 2022-02-09
Improve help message

darkgap:
Following from #166, the README.md have been slightly changed to make more explicit the need of the double dash `--` at the end of the volumes list.

The help message of  the `--volume` command has been slightly changed too, to make it more verbose


tfoote (approved):
Thanks for the clean up.

PR #175, opened 2022-03-23
22.04 tested working

tfoote:


chapulina (approved):
Thanks! `glxgears` works

PR #182, opened 2022-07-04
Set default NVIDIA_DRIVER_CAPABILITIES if it's not set

kenji-miyake:
According to the documentation, the default value of `NVIDIA_DRIVER_CAPABILITIES` is `compute,utility`.
- https://docs.nvidia.com/datacenter/cloud-native/container-toolkit/user-guide.html#driver-capabilities
- https://github.com/NVIDIA/nvidia-container-runtime#nvidia_driver_capabilities

Therefore, `nvidia-smi` can be used for non-NVIDIA images.

```console
$ docker run --rm -it --gpus all ubuntu:22.04 nvidia-smi
Mon Jul  4 10:54:48 2022
+-----------------------------------------------------------------------------+
| NVIDIA-SMI 515.48.07    Driver Version: 515.48.07    CUDA Version: 11.7     |
|-------------------------------+----------------------+----------------------+
| GPU  Name        Persistence-M| Bus-Id        Disp.A | Volatile Uncorr. ECC |
| Fan  Temp  Perf  Pwr:Usage/Cap|         Memory-Usage | GPU-Util  Compute M. |
|                               |                      |               MIG M. |
|===============================+======================+======================|
|   0  NVIDIA GeForce ...  On   | 00000000:01:00.0  On |                  N/A |
| 30%   41C    P0    70W / 290W |   1859MiB /  8192MiB |      0%      Default |
|                               |                      |                  N/A |
+-------------------------------+----------------------+----------------------+

+-----------------------------------------------------------------------------+
| Processes:                                                                  |
|  GPU   GI   CI        PID   Type   Process name                  GPU Memory |
|        ID   ID                                                   Usage      |
|=============================================================================|
+-----------------------------------------------------------------------------+
```

However, `rocker` causes an error because it sets `NVIDIA_DRIVER_CAPABILITIES` to only `graphics` when it's empty.
Our related issue: https://github.com/autowarefoundation/autoware/issues/2452

```console
$ rocker --nvidia --x11 --user ubuntu:22.04
kenji@387c3886ec2d:~$ nvidia-smi
bash: nvidia-smi: command not found
kenji@85ca1235d989:~$ echo $NVIDIA_DRIVER_CAPABILITIES
graphics
```

This PR fixes the behavior.

```console
$ rocker --nvidia --x11 --user ubuntu:22.04
kenji@c6979daf4bad:~$ nvidia-smi
Mon Jul  4 20:01:04 2022
+-----------------------------------------------------------------------------+
| NVIDIA-SMI 515.48.07    Driver Version: 515.48.07    CUDA Version: 11.7     |
|-------------------------------+----------------------+----------------------+
| GPU  Name        Persistence-M| Bus-Id        Disp.A | Volatile Uncorr. ECC |
| Fan  Temp  Perf  Pwr:Usage/Cap|         Memory-Usage | GPU-Util  Compute M. |
|                               |                      |               MIG M. |
|===============================+======================+======================|
|   0  NVIDIA GeForce ...  On   | 00000000:01:00.0  On |                  N/A |
| 30%   41C    P0    70W / 290W |   1874MiB /  8192MiB |      5%      Default |
|                               |                      |                  N/A |
+-------------------------------+----------------------+----------------------+

+-----------------------------------------------------------------------------+
| Processes:                                                                  |
|  GPU   GI   CI        PID   Type   Process name                  GPU Memory |
|        ID   ID                                                   Usage      |
|=============================================================================|
+-----------------------------------------------------------------------------+
kenji@c6979daf4bad:~$ echo $NVIDIA_DRIVER_CAPABILITIES
compute,utility,graphics
```


kenji-miyake:
@tfoote Hello, thank you for developing this tool. Could you take a look at this PR? :pray:

src/rocker/templates/nvidia_snippet.Dockerfile.em
```diff
@@ -25,4 +25,5 @@ COPY --from=glvnd /usr/share/glvnd/egl_vendor.d/10_nvidia.json /usr/share/glvnd/
 
 
 ENV NVIDIA_VISIBLE_DEVICES ${NVIDIA_VISIBLE_DEVICES:-all}
-ENV NVIDIA_DRIVER_CAPABILITIES ${NVIDIA_DRIVER_CAPABILITIES:+$NVIDIA_DRIVER_CAPABILITIES,}graphics
+ENV NVIDIA_DRIVER_CAPABILITIES ${NVIDIA_DRIVER_CAPABILITIES:-compute,utility}
+ENV NVIDIA_DRIVER_CAPABILITIES ${NVIDIA_DRIVER_CAPABILITIES},graphics
```
tfoote, line 29: Thanks for the pointers to the default modes. That looks good to add. Does will this replace #126 for your use cases?

However I'd like to continue to support the ability for the user to manually specify the arguments and then add `,graphics` 

```suggestion
ENV NVIDIA_DRIVER_CAPABILITIES ${NVIDIA_DRIVER_CAPABILITIES:+$NVIDIA_DRIVER_CAPABILITIES,}graphics
```

tfoote (changes requested):
Thanks for sending this in. To keep the change minimal I'd like to consider one change. 

Alternatively, I had a thought that maybe we should just pass `all` by default? It's going to load more of the driver, but I don't see that has having a significant downside, and the user could still reduce the scope by setting it manually.

src/rocker/templates/nvidia_snippet.Dockerfile.em
kenji-miyake: @tfoote Thank you for your comment!

> Does will this replace https://github.com/osrf/rocker/pull/126 for your use cases?

Yes it does because we install CUDA libraries in our image.
See the following links for details.
- https://github.com/autowarefoundation/autoware/blob/ba9d3c4e36d58cbdbd29b85d4b4f54fde001fa0c/docker/autoware-universe/Dockerfile#L29
- https://github.com/autowarefoundation/autoware/blob/ba9d3c4e36d58cbdbd29b85d4b4f54fde001fa0c/setup-dev-env.sh#L126
- https://github.com/autowarefoundation/autoware/blob/ba9d3c4e36d58cbdbd29b85d4b4f54fde001fa0c/ansible/roles/cuda/tasks/main.yaml

> However I'd like to continue to support the ability for the user to manually specify the arguments and then add ,graphics

Sure. I'll commit your suggestion.

kenji-miyake:
> Alternatively, I had a thought that maybe we should just pass all by default? It's going to load more of the driver, but I don't see that has having a significant downside, and the user could still reduce the scope by setting it manually.

@tfoote Although I'm not so familiar with CUDA specs, I personally specifying `all` is acceptable considering the usage of `rocker`.
But in that case, we should set `all` only when it's empty. I mean, for example, `compute,utility,all` is not valid.

```console
$ docker run --rm -it --gpus all -e NVIDIA_DRIVER_CAPABILITIES=compute,utility,graphics ubuntu:22.04
root@f4f80c3232f8:/# exit

$ docker run --rm -it --gpus all -e NVIDIA_DRIVER_CAPABILITIES=compute,utility,all ubuntu:22.04
docker: Error response from daemon: failed to create shim task: OCI runtime create failed: runc create failed: unable to start container process: error during container init: error running hook #0: error running hook: exit status 1, stdout: , stderr: Auto-detected mode as 'legacy'
unsupported capabilities found in 'compute,utility,all' (allowed 'compute,utility'): unknown.
```

tfoote:
Yeah, I think we can then simplify the logic to just set all if it's not previously set. Otherwise it will get whatever is set in the environment.

src/rocker/templates/nvidia_snippet.Dockerfile.em
```diff
@@ -25,4 +25,5 @@ COPY --from=glvnd /usr/share/glvnd/egl_vendor.d/10_nvidia.json /usr/share/glvnd/
 
 
 ENV NVIDIA_VISIBLE_DEVICES ${NVIDIA_VISIBLE_DEVICES:-all}
+ENV NVIDIA_DRIVER_CAPABILITIES ${NVIDIA_DRIVER_CAPABILITIES:-compute,utility}
 ENV NVIDIA_DRIVER_CAPABILITIES ${NVIDIA_DRIVER_CAPABILITIES:+$NVIDIA_DRIVER_CAPABILITIES,}graphics
```
kenji-miyake, line 29: @tfoote Like this?

```suggestion
ENV NVIDIA_DRIVER_CAPABILITIES ${NVIDIA_DRIVER_CAPABILITIES:-all}
```

tfoote (approved):
Thanks for iterating it's small and simple now.

Issue #183, opened 2022-07-08
Add option to not pass host user $SHELL to container when using --user?


rcywongaa:
I use `zsh` in my host environment but `zsh` isn't installed in my container.  It seems with `--user`, `rocker` automatically sets the container shell to use the host user shell

https://github.com/osrf/rocker/blob/28a8e796ed7fe9b450a80a6385b3a2f56e85e3d0/src/rocker/templates/user_snippet.Dockerfile.em#L19

This causes the following warning to pop up in various places
```
Warning: missing or non-executable shell '/bin/zsh'
```

Given that a container usually has a stripped down version of the host environment, it doesn't seem like a good idea to assume the container will have/use the same shell as the host.  

Should there be an option to disable this shell setting?

tfoote:
Yeah, an option like `--user-override-shell` might make sense similar to how you can override the username

https://github.com/osrf/rocker/blob/28a8e796ed7fe9b450a80a6385b3a2f56e85e3d0/src/rocker/extensions.py#L233-L236

rcywongaa:
Cool!  I can work on this and submit my first PR :D

Should the default behavior (in the absence of `--user-override-shell`) be changed to use the container's default shell then, i.e. only use a different shell when explicitly provided ?

PR #185, opened 2022-07-09
Make using user shell optional

rcywongaa:
Resolves https://github.com/osrf/rocker/issues/183

Adds option `--use-user-shell` and changes default behavior to not override container shell

Let me know if you disagree with this change in default behavior and I'll amend this to implement `--user-override-shell` instead

### DIFF
--- a/src/rocker/extensions.py
+++ b/src/rocker/extensions.py
@@ -222,6 +222,11 @@ class User(RockerExtension):
             substitutions['dir'] = os.path.join('/home/', cliargs['user_override_name'])
         substitutions['user_preserve_home'] = True if 'user_preserve_home' in cliargs and cliargs['user_preserve_home'] else False
         substitutions['home_extension_active'] = True if 'home' in cliargs and cliargs['home'] else False
+        if 'user_override_shell' in cliargs and cliargs['user_override_shell'] is not None:
+            if cliargs['user_override_shell'] == '':
+                substitutions['shell'] = None
+            else:
+                substitutions['shell'] =  cliargs['user_override_shell']
         return em.expand(snippet, substitutions)
 
     @staticmethod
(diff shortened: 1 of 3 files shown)


src/rocker/extensions.py
```diff
@@ -238,6 +239,10 @@ def register_arguments(parser, defaults={}):
             action='store_true',
             default=defaults.get('user-preserve-home', False),
             help="Do not delete home directory if it exists when making a new user.")
+        parser.add_argument('--use-user-shell',
```
tfoote, line 242: I proposed `--user-override-shell` 

The beginning of the argument matches the plugin name which I definitely want. All arguments relating to that plugin should start with `--user` for grouping and visibility.

tfoote (changes requested):
Thanks for putting together the PR for this.

There's definitely a case to be made that the default could be the default shell inside the container versus the default shell from the outside. In the cases that I've been supporting they've usually been the same. But for the use cases when I'm trying to replicate the user's development experience inside the container reproducing their shell inside is useful too. (Aka you use `--home` with `--user` you get you whole shell environment with all your settings etc.) As this is the primary use case I've been targeting I'd like to keep that as the default. 

With the way you've laid out the logic there's currently only two possible shells. Either the user's external shell or the internal default. Whereas I was proposing you have the user's external shell, or alternatively you can specify a specific shell which could be any shell you want. This mirrors the `override-username` where you either get the external username or you can force any username you want based on knowledge of your target. 

In general if you're using zsh or some other shell externally I would suggest that you consider also making a small extension which will inject zsh into the container you're using that way you can have your preferred shell too when you use rocker. The other possible extension would be to have a `--user-no-set-shell` option. Or it maybe could be not set if the override option is explicitly set to an empty string. 

In the larger picture I'd like to get to having some rocker config file format to help people keep their command lines less verbose and set defaults for their operations.

tfoote:
Oh sorry I also forgot to mention that I'd like to also make sure to add coverage of this extra option in the unit tests before merging.

rcywongaa:
My (probably incorrect) use case of `--user` is to simply avoid having to manually create a non-root user so that any files generated within the container ends up having the same owner as the host.  Hence my settings were more towards create the user and nothing more.  My zsh is heavily customized with many plugins which probably doesn't make sense to replicate in the container.

Anyway, I understand the original intention.  I guess most container's default / available shell is well known (`bin/sh`) anyway

tfoote (approved):
Thanks for the updates. It looks good to me. 

Your use case isn't incorrect. That is a primary function of the `--user` option, it's more of a prioritization of the different use cases.

Issue #187, opened 2022-07-20
Cannot run a persistent container with x11


tonynajjar:
Steps to repoduce:

```
rocker --nocleanup --x11 --name test osrf/ros:rolling-desktop 
exit
docker start test
```

The error:
```
Error response from daemon: failed to create shim task: OCI runtime create failed: runc create failed: unable to start container process: error during container init: error mounting "/tmp/.dockerj054ncrt.xauth" to rootfs at "/tmp/.dockerj054ncrt.xauth": mount /tmp/.dockerj054ncrt.xauth:/tmp/.dockerj054ncrt.xauth (via /proc/self/fd/7), flags: 0x5000: not a directory: unknown: Are you trying to mount a directory onto a file (or vice-versa)? Check if the specified host path exists and is the expected type
Error: failed to start containers: test
```

tonynajjar:
I think this is still an issue when the system is rebooted, everything is tmp is deleted :/

tfoote:
Yeah, that definitely happens. I don't want to be generating a lot of persistent files in the default case. But maybe we can push the temp files into an alternative location. I was originally thinking of adding an argument but the python implementation already has an environment variable (apparently 3 actually) for this case that you could use.

https://docs.python.org/3/library/tempfile.html#tempfile.mkstemp

`The default directory is chosen from a platform-dependent list, but the user of the application can control the directory location by setting the TMPDIR, TEMP or TMP environment variables. `

PR #188, opened 2022-07-20
Remove tempfile file generator

tonynajjar:
Fixes https://github.com/osrf/rocker/issues/187

### DIFF
--- a/src/rocker/nvidia_extension.py
+++ b/src/rocker/nvidia_extension.py
@@ -12,11 +12,9 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
-import grp
 import os
 import em
 import getpass
-import tempfile
 from packaging.version import Version
 import pkgutil
 from pathlib import Path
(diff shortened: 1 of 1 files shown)


src/rocker/nvidia_extension.py
```diff
@@ -42,18 +40,18 @@ def get_name():
     def __init__(self):
         self.name = X11.get_name()
         self._env_subs = None
-        self._xauth = tempfile.NamedTemporaryFile(prefix='.docker', suffix='.xauth')
```
tonynajjar, line 45: There is probably a reason behind this?

tfoote: This was made unique to keep multiple containers from interfering with each other.

The tempfile is the safe way to make a file like this to ensure that there's no collisions etc.

What I would recommend instead is to extend the nocleanup logic to make the tempfile persist. 

An extension would be to also think about how to have a second step which would clean up any dangling resources like this for a container previously passed nocleanup

tfoote:
Could you take a look at #189 as an alternative?

tonynajjar:
Hey, sorry for the late reply, I'll test it out by the end of the week

tonynajjar:
Works like a charm, closing this PR

PR #190, opened 2022-07-30
Added note for linking Intel Xe cards with rocker

zmk5:
When running on an 11th gen Intel processor, I continually got a `(Segmentation Fault)` and `Forcing OpenGl version 0.` error when launching Rviz or any other GUI that required accelerated graphics within a container. This only happened when I linked the integrated graphics with the `--devices /dev/dri/card0` option as suggested by the documentation.

Further research showed that the `/dev/dri/renderD128` is the correct device to link for Intel chipsets with Xe graphics. Therefore, I've added an extra note on the `rocker` README for future users.

Please let me know if there is anything else I should change!

### DIFF
--- a/README.md
+++ b/README.md
@@ -42,16 +42,10 @@ Note, that changing this setting will lead to a `Failed to initialize NVML: Unkn
 
 ## Intel integrated graphics support
 
-For intel integrated graphics support you will need to mount through a specific device
+For [Intel integrated graphics support](https://www.intel.com/content/www/us/en/develop/documentation/get-started-with-ai-linux/top/using-containers/using-containers-with-the-command-line.html) you will need to mount the `/dev/dri` directory as follows:
 
 ```
---devices /dev/dri/card0
-```
-
-For Intel Xe integrated graphics, you will need to use the `renderD128` option instead:
-
-```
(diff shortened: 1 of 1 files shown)


tfoote (approved):
Thanks for the contribution. Do you have any of the links that you used to find this device name that could potentially be added at least to this thread for future reference?

zmk5:
Hey @tfoote. Doing more of a deep dive, I found [this](https://www.intel.com/content/www/us/en/develop/documentation/get-started-with-ai-linux/top/using-containers/using-containers-with-the-command-line.html) documentation by Intel, and they suggest just connecting the integrated graphics with `--device=/dev/dri` instead of specifying the specific card. Maybe we should change the documentation to just do that instead?

I've checked it on my system and using rocker with `--device /dev/dri` works!

tfoote:
That's neat, I didn't realize you could do the whole directory. Can you update the instructions to suggest that instead of having the two specific callouts?

zmk5:
Of course. I've updated the documentation and linked the official Intel documentation for future users as well.

tfoote (approved):
Great. Thanks for the updates!

Issue #176, opened 2022-04-28
unable to map ports


Linusnie:
Hi, I find rocker very handy for mapping user privileges, enabling GUIs etc. But as far as I can tell port mapping (e.g. -p 5050:5050) doesn't work, so I have to switch back to regular docker in some situations.

Is it possible to do port mapping somehow with rocker? And if not could it be added to the interface?

jaron-l:
Caveat: I've never used this package before nor know much about it but...

I believe you are looking for [run arg mappings](https://github.com/sloretz/off-your-rocker#docker-run-args).

ErikKrauter:
I have the same issue. Event after the PR #201. I cannot use -p or --port with rocker.

tfoote:
This should be covered by #201 and OYR run arg mappings. @ErikKrauter if you are still having an issue please open a new issue with instructions for how to reproduce.

PR #201, opened 2022-09-19
Add port and expose extensions with tests

willcbaker:
Add support for using -p and --expose with container
https://docs.docker.com/engine/reference/commandline/run/#publish-or-expose-port--p---expose


tfoote (approved):
Thanks for the additions with good coverage.

tfoote:
Fixes #176

PR #205, opened 2022-10-26
Removed '-v' for rocker's version. Only --version should be available.

gstavrinos:
The reasoning between this change is that it creates a conflict with the -v flag of the `docker volume` extension.


gstavrinos:
For a quick test of the current (problematic) behaviour just run:
`rocker -v`
The output (at least in my case) is:
```
rocker 0.2.10
```

The first line comes from the `docker volume` extension and the second one is the actual rocker version.

tfoote (approved):
Thanks yes, this is good to deconflict. The shorthand for version is fine to drop for this benefit.

PR #218, opened 2023-04-04
Generalize username as build argument

orensbruli:
This makes it easier to reuse the generated Dockerfile in different projects.

### DIFF
--- a/src/rocker/templates/user_snippet.Dockerfile.em
+++ b/src/rocker/templates/user_snippet.Dockerfile.em
@@ -1,4 +1,5 @@
 ARG USERNAME=@(name)
+ARG HOMEDIR=@(dir)
 
 # make sure sudo is installed to be able to give user sudo access in docker
 RUN if ! command -v sudo >/dev/null; then \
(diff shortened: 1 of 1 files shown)


src/rocker/templates/user_snippet.Dockerfile.em
```diff
@@ -1,3 +1,6 @@
+ARG USERNAME=@(name)
+ARG HOMEDIR=@(dir)
+
```
tfoote, line 3: These are not the only parameters to the Dockerfile, if we're doing this we should cover them all. The uid and gid are also important if you change the user.

tfoote (changes requested):
I understand your idea. There's at least a few other arguments that should be exposed. 

Can you explain a little bit more about your use case? It would be great to understand it alittle bit more. In general rocker has been designed to regenerate the overlayed dockerfiles at runtime because they are specific to the runtime user not the build time user.  If you have the ability to manually predict the runtime user that's fine but it's going to be fragile if someone else tries to reuse your image generated as such. 

One of my thoughts in the future was to allow caching of the layers. But to that end they need to embed all the parameters into the cach invalidation such that rocker can know if the cached image matches the current runtime environment. How would you think about this working in that context.

src/rocker/templates/user_snippet.Dockerfile.em
orensbruli: My specific user case was to quickly create my own Dockerfile for another project and after seeing the dumped one in the rocker output I guessed it would be a good addition to other people.
It would be extended for many common use cases in which we can't use rocker in a docker compose but we could create our own images with the dockerfiles generated by rocker. 

I really think that the Dockerfiles that rocker internally uses are a nice reference for other projects needing nvidia, user, x11, etc in it projects.

Maybe I could implement an option in rocker to save the Docker files instead of building itself.

tfoote: The main design goal of rocker is to provide dynamic content that is not possible to capture in a container image distributed by a project. (Aka it's not appropriate to bake in the $USERNAME to an image. Which is exactly why rocker detects and generates the content dynamically. I have ideas about how to enable baking in cached content, but I think that trying to make the dynamically generated Dockerfiles exportable will just lead to people hitting unexpected errors. For example as you've proposed this will break unless you happen to have the same UID.

orensbruli: I understand. Thank you so much for taking the time to explain it. Best!

Issue #35, opened 2019-02-14
display build errors


tfoote:
When there's a build error the output in rocker is suppressed. 

As you can see manually running the command afterwards shows that it's failing due to executable file not found. But the rocker build just ends with an error code and no information. It is an invalid configuration but that doesn't mean we shouldn't show the error message. 

```
$ rocker --home --user hello-world --noex
Plugins found: ['dev_helpers', 'home', 'nvidia', 'pulse', 'user']
Active extensions ['home', 'user']
Writing dockerfile to /tmp/tmpwfb8q_7_/Dockerfile
vvvvvv
# Preamble from extension [home]

# Preamble from extension [user]


FROM hello-world
# Snippet from extension [home]

# Snippet from extension [user]
# make sure sudo is installed to be able to give user sudo access in docker
RUN apt-get update \
 && apt-get install -y \
    sudo \
 && apt-get clean

RUN useradd -U --uid 1000 -ms /bin/bash osrf \
 && echo "osrf:osrf" | chpasswd \
 && adduser osrf sudo \
 && echo "osrf ALL=NOPASSWD: ALL" >> /etc/sudoers.d/osrf
# Commands below run as the developer user
USER osrf


^^^^^^
Building docker file with arguments:  {'path': '/tmp/tmpwfb8q_7_', 'rm': True, 'decode': True, 'nocache': False, 'tag': 'rocker_hello-world_home_user'}
building > Step 1/4 : FROM hello-world
building >  ---> fce289e99eb9
building > Step 2/4 : RUN apt-get update  && apt-get install -y     sudo  && apt-get clean
building >  ---> Running in 6f95c70df683
2 osrf@osrf-toughpad1:~$ docker run -ti --rm 6f95c70df683
Unable to find image '6f95c70df683:latest' locally
docker: Error response from daemon: pull access denied for 6f95c70df683, repository does not exist or may require 'docker login'.
See 'docker run --help'.
125 osrf@osrf-toughpad1:~$ docker run -ti --rm fce289e99eb9 bash
docker: Error response from daemon: OCI runtime create failed: container_linux.go:344: starting container process caused "exec: \"bash\": executable file not found in $PATH": unknown.
127 osrf@osrf-toughpad1:~$ docker run -ti --rm fce289e99eb9 sh
docker: Error response from daemon: OCI runtime create failed: container_linux.go:344: starting container process caused "exec: \"sh\": executable file not found in $PATH": unknown.
[...]

PR #227, opened 2023-05-04
Issue - #35 - Error display when docker-build perform

BabaYaga1221:
The issue of suppressed output when encountering build errors in the docker_build function has been resolved by modifying the code to handle errors by capturing and printing the output message instead of suppressing it. The modified code now allows users to see the full error message and easily diagnose the issue.
@tfoote

### DIFF
--- a/src/rocker/core.py
+++ b/src/rocker/core.py
@@ -135,9 +135,11 @@ def get_docker_client():
 
 def docker_build(docker_client = None, output_callback = None, **kwargs):
     image_id = None
+    build_success = False
 
     if not docker_client:
         docker_client = get_docker_client()
+
     kwargs['decode'] = True
     for line in docker_client.build(**kwargs):
         output = line.get('stream', '').rstrip()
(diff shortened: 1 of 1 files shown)


tfoote (changes requested):
Thanks for looking at this. However it gives the code backtrace from the code path and not the error from the underlying command that I'd like to expose. 

You can see if you run `rocker --user hello-world --noex` that it still prints `Exception: Build failed: no more output and success not detected` with the backgrace

What it should print is the error from the build step.
If you run it after the above and substitute the last image id into this command `docker run -ti --rm 4939e01e77c4` 

You'll see the useful error: `docker: Error response from daemon: unable to find user root: no matching entries in passwd file.` Which is what should be visible to the user as that's the underlying error not the backtrace.

BabaYaga1221:
@tfoote Would you like to review the code to see if the build errors can be traced back more clearly?

tfoote (changes requested):
It does print the build commands. 
```
Exception: Build failed: Step 1/7 : FROM hello-world
 ---> feb5d9fea6a5
Step 2/7 : USER root
 ---> Using cache
 ---> 4939e01e77c4
Step 3/7 : RUN if ! command -v sudo >/dev/null; then       apt-get update       && apt-get install -y sudo       && apt-get clean;     fi
 ---> Running in 5a5af0f73975
```

But it doesn't print the underlying error so that the user can know what failed. 

And secondly, just throwing the exception and printing the build output is not a good user experience. The exception should be caught and helpful context should be supplied to the user about what was being tried to run, what the error was, and hopefully how they can reproduce it for debugging.

BabaYaga1221:
hi @tfoote, I apologize for being inactive on this PR, but, can you please help to what should I change, 
last time when I work on it, I only ended only one solution that I should manually describe the error and provide particular steps to resolve.

tfoote:
You need to catch the underlying build error output. And then present what was attempted as well as the error output for the user to understand what failed so they can then fix the problem themselves.

PR #234, opened 2023-05-12
fix: nvidia arg

### LINKED ISSUE #233: nvidia incorrectly enabled even if not selected
There's a regression in #211 @amadeuszsz 

With that merged the --nvidia plugin is always activated even if there's no arguments passed. 

```
$ rocker ubuntu | head
Active extensions ['nvidia']
```

amadeuszsz:
Fixes #233

### DIFF
--- a/src/rocker/nvidia_extension.py
+++ b/src/rocker/nvidia_extension.py
@@ -138,7 +138,7 @@ class Nvidia(RockerExtension):
             choices=['auto', 'runtime', 'gpus'],
             nargs='?',
             const='auto',
-            default=defaults.get(Nvidia.get_name(), 'auto'),
+            default=defaults.get(Nvidia.get_name(), None),
             help="Enable nvidia. Default behavior is to pick flag based on docker version.")
 
 class Cuda(RockerExtension):


src/rocker/nvidia_extension.py
```diff
@@ -138,7 +138,6 @@ def register_arguments(parser, defaults={}):
             choices=['auto', 'runtime', 'gpus'],
             nargs='?',
             const='auto',
-            default=defaults.get(Nvidia.get_name(), 'auto'),
```
tfoote, line 141: I think that we need to have a None as the default. Removing this capability prevents the use of the `defaults = {}` map. We basically need a weird state where there's effectively a 4th state which is None. Then if you set --nvidia you have the ternary selection you've defined. The way we've done this in the past is to use a second argument and I think that might make sense to follow that pattern here too. The option `--nvidia` enables the plugin. And use `--nvidia-runtime=[auto, runtime, gpus]` and defaults to `auto`. It requires you to pass both argument to override the runtime, but very few people will want to do that so this will be simpler on average, but enable the power users like yourself.

tfoote (changes requested):

src/rocker/nvidia_extension.py
amadeuszsz: > I think that we need to have a None as the default. 

Before I start could you tell me if the 2nd fix meets package pattern? I'm just testing `test_nvidia.NvidiaTest.test_nvidia_extension_basic` and all kwargs combinations and it's working with default argument now. Otherwise, we will do as you explained.

tfoote (approved):
Interestingly that does actually work. I was not expecting argparse to be ok if the default value was outside of the choices. 

Thanks for the cleanup.

Issue #37, opened 2019-02-25
Add the ability for extensions to have requirements on other extension


tfoote:
For example the nvidia extension will need the x11 extension to be useful.

Also the home extension should require the user extension for permissions.

There' s a question of ordering, required before or not. And for each sequential element aka snippets.

Preamble elements are named so order independent, and arguments are order independent as far as I know as long as they're escaped properly.

This will resolve this todo: https://github.com/osrf/rocker/blob/cea068015fd7e1026d3887c103a71d86557b5dbb/test/test_nvidia.py#L134

stonier:
Experimenting with this at https://github.com/stonier/groot_rocker/pull/31.

Is it in the vicinity of what you were looking for?

agyoungs:
This looks stale, but would be pretty useful for a few things I'm working on. Is there any progress on this? I'm happy to take a stab at it if the solution provided by @stonier isn't quite complete and they don't have the time to work on it at the moment.

PR #243, opened 2023-09-22
Separate out pre and post user snippets.

tfoote:
Regular snippets can assume root access.
Post snippets can assume USER is set to the final user.

To remove special logic needed in #242 

I had thought that this would resolve #130 however the assumption in #37 "arguments are order independent as far as I know as long as they're escaped properly." Environment variables are order dependent. 

I think that the solution for that is likely to encourage plugins to set ENV values via snippet versus command line. Though that also has runtime vs build time implications.

Todos
* [ ] add coverage of the new snippets

### DIFF
--- a/src/rocker/core.py
+++ b/src/rocker/core.py
@@ -15,6 +15,7 @@
 from collections import OrderedDict
 import io
 import os
+import pwd
 import re
 import sys
 
@@ -62,6 +63,11 @@ class RockerExtension(object):
         return ''
 
     def get_snippet(self, cliargs):
+        """ Get a dockerfile snippet to be executed as ROOT."""
+        return ''
+
+    def get_user_snippet(self, cliargs):
+        """ Get a dockerfile snippet to be executed after switchingto the expected USER."""
         return ''
 
     def get_files(self, cliargs):
@@ -132,6 +138,9 @@ def get_docker_client():
             ' This is usually by being a member of the docker group.'
             ' The underlying error was:\n"""\n%s\n"""\n' % ex)
 
+def get_user_name():
+    userinfo = pwd.getpwuid(os.getuid())
+    return getattr(userinfo, 'pw_' + 'name')
 
 def docker_build(docker_client = None, output_callback = None, **kwargs):
     image_id = None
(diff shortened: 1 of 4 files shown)


src/rocker/core.py
```diff
@@ -338,14 +347,27 @@ def write_files(extensions, args_dict, target_directory):
...
         dockerfile_str += el.get_snippet(args_dict) + '\n'
+    # Set USER if user extension activated
+    if 'user' in args_dict and args_dict['user']:
+        if 'user_override_name' in args_dict and args_dict['user_override_name']:
+            username = args_dict['user_override_name']
+        else:
+            username = get_user_name()
+        dockerfile_str += f'USER {username}\n'
+    # USER snippets
+    for el in extensions:
```
agyoungs, line 368: These user snippets will execute whether the `user` extension is present or not. Which I think is fine, however, some extensions might expect there to be a `/home/<user>` directory at this point. Is there a good way for those extensions to detect and handle that case?

I suppose one way (albeit cumbersome) is to add a `user_exists_preamble` and a `no_user_preamble` that gets called immediately before the `user_snippet`. This allows an extension to set up variables as needed to execute the user snippet as root or as a user.

agyoungs (commented):
I like the flexibility this offers to extensions. I just had one comment regarding the user_snippet.

src/rocker/core.py
agyoungs: I realized since I posted this that the `get_user_snippet` method has the cli_args available. An extension should be able to determine if the `user` extension is enabled/disabled within this method call and can modify the docker return string as needed.

src/rocker/core.py
```diff
@@ -338,14 +347,27 @@ def write_files(extensions, args_dict, target_directory):
...
     for el in extensions:
         dockerfile_str += '# Snippet from extension [%s]\n' % el.name
         dockerfile_str += el.get_snippet(args_dict) + '\n'
+    # Set USER if user extension activated
+    if 'user' in args_dict and args_dict['user']:
+        if 'user_override_name' in args_dict and args_dict['user_override_name']:
+            username = args_dict['user_override_name']
+        else:
+            username = get_user_name()
+        dockerfile_str += f'USER {username}\n'
```
agyoungs, line 366: Should we consider adding `USER <user>` string in front of each extension call? Just in case the extension calls `USER root` in the user snippet? Or should that case be ignored since that would be non-standard

tfoote: That would be non-standard and also add a lot of potential layers to each build which is something that could be a limitation. It would be good to add docs that the snippets are expected not to change the USER.

tfoote:
Merge with #242 This adds the user resolution that was simplified out of that one.

PR #244, opened 2023-09-25
Create directory structure when writing files

agyoungs:
This simple change will allow an extension to have a directory structure for the files to copy into the container. This can simplify structuring the directory in the container if you simply wish to preserve the structure from the host machine.

### DIFF
--- a/src/rocker/core.py
+++ b/src/rocker/core.py
@@ -325,12 +325,12 @@ class DockerImageGenerator(object):
 def write_files(extensions, args_dict, target_directory):
     all_files = {}
     for active_extension in extensions:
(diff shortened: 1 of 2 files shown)


tfoote (changes requested):
This seems reasonable. This took longer to understand as looks like you made a few unnecessary changes and it only needs to be the one line change in L334. 

Can you also add some tests to cover different cases with this as well as a proposed documentation update too?

agyoungs:
I forgot to come back to this until now, but let me know if you think the additional test is sufficient. I also added an explicit check for paths outside the temp directory.

tfoote (approved):
Thanks for adding the tests.

PR #250, opened 2023-11-13
Allow binary files to be copied into the build context

agyoungs:
The current write routine will not work if the data is a `bytes` type. This can happen if the `get_files` function reads in a file using the `b` mode (appropriate if it's a binary file).

This change simply checks if the data is a `bytes` class and sets the write mode accordingly


agyoungs:
> Thanks, this looks good. It would be great to have a quick unit test before merging to make sure there's not a regression in the future. I'll go over this once more when I'm back in the office next week.

I've now added another file for the file writing test that should cover this.

tfoote (approved):
Thanks for adding the test.

Issue #257, opened 2023-11-29
libGL error: MESA-LOADER: failed to retrieve device information


woensug-choi:
On freshly install Ubuntu 22.04 Jammy LTS. Without doing anything,
I've installed rocker with,
```bash
pip3 install rocker
pip3 install --force-reinstall git+https://github.com/osrf/rocker.git@main
rocker --version
# rocker 0.2.12
```
and ran Example in README
```bash
rocker --nvidia --x11 osrf/ros:noetic-desktop-full gazebo
```

and Got error saying
```
libGL error: MESA-LOADER: failed to retrieve device information
```

I was able to fix the problem by adding `--volume /dev:/dev` in rocker argument. which adds `-v /dev:/dev` to docker argument.
```bash
rocker --volume /dev:/dev --nvidia --x11 osrf/ros:noetic-desktop-full gazebo
```

tfoote:
On your fresh install. Do you have the NVIDIA drivers installed And you should make sure that you've installed and setup nvidia-docker or now the NVIDIA Container Toolkit: https://docs.nvidia.com/datacenter/cloud-native/container-toolkit/latest/install-guide.html ?

woensug-choi:
Yes. I had toolkit installed.

tfoote:
As far as I can tell your device isn't being mounted correctly. Your solution of mounting all of /dev tells me that the device is available. It's just a matter of understanding what your device is and making sure to mount it. Mounting all devices is too broad a brush. More specific feedback in #258 which I closed as too broad a solution. But with a more targeted fix we could add a solution.

noah-curran:
I'll preface this that I'm a Docker noob so its entirely possible I'm doing something wrong... but I've been having this issue with my fresh install as well. I've spent several hours over the past 5 days on this so I've ruled out many of the common points of advice and came here to discover this chain.

Similar to @woensug-choi I'm using Ubuntu 22.04 and Nvidia driver version 535. I have an RTX 4070 and understand that 535 is not a tested driver version, but my GPU does not support the maximum driver version tested, 470.

[...]

Issue #206, opened 2022-11-09
rviz2 RenderSystem errors inside the autoware-container on ubuntu 20.04 on nvidia GPU


AndriiChumak145:
Environment: 
- Ubuntu 20.04, x86-6 
- nvidia GPU with Driver Version: 520.61.05    CUDA Version: 11.8.
- rocker 0.2.10

I have run an autoware container as described in their docs:
`rocker --nvidia --x11 --user --volume $HOME/autoware --volume $HOME/autoware_map -- ghcr.io/autowarefoundation/autoware-universe:latest-cuda`
After I tried to use rviz2 I received the following output:
```
[ERROR] [1667991076.248236751] [rviz2]: Unable to create the rendering window after 100 tries
terminate called after throwing an instance of 'std::runtime_error'
  what():  Unable to create the rendering window after 100 tries
Aborted (core dumped)

```
After that I tried to run your example:
`rocker --nvidia --x11 osrf/ros:crystal-desktop rviz2`
and rviz window was displayed. However, I got the following libGL errors and as the output suggests it was trying to load intel drivers instead of nvidia:
```
QStandardPaths: XDG_RUNTIME_DIR not set, defaulting to '/tmp/runtime-root'
libGL error: MESA-LOADER: failed to retrieve device information
libGL error: Version 4 or later of flush extension not found
libGL error: failed to load driver: i915
libGL error: failed to open drm device: No such file or directory
libGL error: failed to load driver: iris
libGL error: MESA-LOADER: failed to retrieve device information
libGL error: Version 4 or later of flush extension not found
libGL error: failed to load driver: i915
libGL error: failed to open drm device: No such file or directory
libGL error: failed to load driver: iris
[INFO] [rviz2]: Stereo is NOT SUPPORTED
[INFO] [rviz2]: OpenGl version: 3.1 (GLSL 1.4)
[INFO] [rviz2]: Stereo is NOT SUPPORTED
```

I also managed to run rviz2 without errors without nvidia but the performance was unsatisfactory.

tfoote:
I cannot reproduce your issue. There might be something in your environment or your volumes selecting the intel driver?


Here's my attempt I skipped the volumes as I don't have those setup. 
<details>

```
🟢 SUCCESS] ❯ rocker --nvidia --x11 --user -- ghcr.io/autowarefoundation/autoware-universe:latest-cuda
Extension volume doesn't support default arguments. Please extend it.
Active extensions ['nvidia', 'x11', 'user']
[...]

PR #258, opened 2023-11-29
Add `-v /dev:/dev` to  X11 default argument to fix libGL error: MESA-LOADER

woensug-choi:
On freshly install Ubuntu 22.04 Jammy LTS. Without doing anything,
I've installed rocker with,
```bash
pip3 install rocker
pip3 install --force-reinstall git+https://github.com/osrf/rocker.git@main
rocker --version
# rocker 0.2.12
```
and ran Example in README
```bash
rocker --nvidia --x11 osrf/ros:noetic-desktop-full gazebo
```

and Got error saying
```
libGL error: MESA-LOADER: failed to retrieve device information
```

I was able to fix the problem by adding `--volume /dev:/dev` in rocker argument. which adds `-v /dev:/dev` to docker argument.
```bash
rocker --volume /dev:/dev --nvidia --x11 osrf/ros:noetic-desktop-full gazebo
```

I believe the right position to add `-v /dev:/dev` is `--x11` argument tag since it wouldn't break even if /dev doesn't exist.

Related articles
https://github.com/osrf/rocker/issues/257
https://github.com/osrf/rocker/issues/206
https://github.com/kinu-garage/hut_10sqft/issues/819

### DIFF
--- a/src/rocker/nvidia_extension.py
+++ b/src/rocker/nvidia_extension.py
@@ -49,7 +49,7 @@ class X11(RockerExtension):
         return "  -e DISPLAY -e TERM \
   -e QT_X11_NO_MITSHM=1 \
   -e XAUTHORITY=%(xauth)s -v %(xauth)s:%(xauth)s \
-  -v /tmp/.X11-unix:/tmp/.X11-unix \
+  -v /tmp/.X11-unix:/tmp/.X11-unix -v /dev:/dev \
   -v /etc/localtime:/etc/localtime:ro " % locals()
 
     def precondition_environment(self, cliargs):


tfoote:
That's an improperly broad solution for getting nvidia to work. In particular you say this is on a fresh installation. Do you have your NVIDIA drivers setup and docker-nvidia setup as well? The fact that it's trying to use MESA suggest that you it's not detecting NVIDIA. And the fix for a non-NVIDIA gpu is usually to use `--device /dev/dri/card0` I would suggest that you try that instead of mounting the full `/dev` If that doesn't fix it you should find out the specific device that you need.

tfoote (changes requested):
I have double checked my system and cannot reproduce. And if there is something missing, this is way too broad of a solution to be appropriate.

woensug-choi:
> That's an improperly broad solution for getting nvidia to work. In particular you say this is on a fresh installation. Do you have your NVIDIA drivers setup and docker-nvidia setup as well? The fact that it's trying to use MESA suggest that you it's not detecting NVIDIA. And the fix for a non-NVIDIA gpu is usually to use `--device /dev/dri/card0` I would suggest that you try that instead of mounting the full `/dev` If that doesn't fix it you should find out the specific device that you need.
> 
> 
> 
> 
> 
> 

I will try --device /dev/dri/card0
meanwhile, I did install NVIDIA driver and toolkit installed. nvidia-smi shows fine. Also glxinfo shows no particular error message. glxgears works fast.

woensug-choi:
Yes, `--device /dev/dri/card0` solves problem! but without it I get the same error. Does this mean that we still need a PR to add `--device /dev/dri/card0`?

woensug-choi:
Hmm. More notes,

both with and without `--device /dev/dri/card0`, when running `nvidia-smi` inside the container does get correct NVIDIA GPU. But opening gazebo doesn't work without `--device /dev/dri/card0`

tested with,
```bash
rocker --nvidia --x11 osrf/ros:noetic-desktop-full /bin/bash
nvidia-smi
```
prints
```bash
Wed Nov 29 20:51:02 2023       
+---------------------------------------------------------------------------------------+
| NVIDIA-SMI 535.129.03             Driver Version: 535.129.03   CUDA Version: 12.2     |
|-----------------------------------------+----------------------+----------------------+
| GPU  Name                 Persistence-M | Bus-Id        Disp.A | Volatile Uncorr. ECC |
| Fan  Temp   Perf          Pwr:Usage/Cap |         Memory-Usage | GPU-Util  Compute M. |
|                                         |                      |               MIG M. |
|=========================================+======================+======================|
|   0  Quadro RTX 3000 with Max...    Off | 00000000:01:00.0 Off |                  N/A |
| N/A   53C    P0              22W /  65W |      5MiB /  6144MiB |      0%      Default |
|                                         |                      |                  N/A |
+-----------------------------------------+----------------------+----------------------+
                                                                                         
+---------------------------------------------------------------------------------------+
| Processes:                                                                            |
|  GPU   GI   CI        PID   Type   Process name                            GPU Memory |
|        ID   ID                                                             Usage      |
|=======================================================================================|
+---------------------------------------------------------------------------------------+
```
when running gazebo,
```bash
gazebo
# libGL error: MESA-LOADER: failed to retrieve device information
```

It's a fresh install of Ubuntu 22.04 with recommended NVIDIA driver installation which were selected automatically during installation process.

tfoote:
>  I thought gpus -all whould deal with this.

My understanding is that the `--gpus all` only catches discrete GPUs I believe. /dev/dri/card0 is the integrated intel graphics If you're using that I don't believe that you're using the NVIDIA card.

> with recommended NVIDIA driver installation

This may be a generic ubuntu recommendation not necessarly what you want

I see that you're using the 535 driver. I have only ever tested up to the 470 NVIDIA driver. (See the README) There's a level of required compatibility between the internal and external drivers and going to the 

Are you using the 535-open drivers? https://forums.linuxmint.com/viewtopic.php?t=401149

I see reported issues with 535 reported here too: https://github.com/NVIDIA/nvidia-docker/issues/1767

woensug-choi:
The `535` NVIDIA driver wasn't `open` driver version apart from other `open` versions of drivers in `Additional Drivers` in ubuntu 22.04. I've tested with `470.223.02` the proprietary version of NVIDIA driver (I did reboot by the way), but also didnt' work.

tfoote:
I'm not sure what I can do to help you. I can't reproduce your issue. Does Gazebo run on the host machine with NVIDIA support?

tfoote:
https://github.com/containers/podman/issues/7801#issuecomment-722574489

I found another pattern of potential devices that might be a solution instead of `/dev/dri/card0` `/dev/dri/renderD128` Also I see that the p16s can also have nvidia chips so that might mount up in a different location.  Because this is too generic I don't want to merge it as proposed so I'm going to close this. But if there's a more specific device that we can detect and mount if needed that would be a good reason to reopen this.

PR #268, opened 2024-02-26
Add a shorthand option for switching to alternate rmws

tfoote:
This adds a `--rmw <IMPL>` option to rocker to enable selecting the default RMW layer. 

Remaining
* [x] Add some unit tests
* [x] Consider a refactor to host this externally from the core

### DIFF
--- a/test/test_rmw_extension.py
+++ b/test/test_rmw_extension.py
@@ -51,7 +51,6 @@ class rmwExtensionTest(unittest.TestCase):
         
 
         mock_cliargs = {'rmw': ['cyclonedds']}
-        # TODO self.assertEqual(p.get_snippet(mock_cliargs), '')
         self.assertEqual(p.get_preamble(mock_cliargs), '')
         args = p.get_docker_args(mock_cliargs)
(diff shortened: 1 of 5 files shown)


test/test_rmw_extension.py
```diff
@@ -0,0 +1,95 @@
...
+        plugins = list_plugins()
+        rmw_plugin = plugins['rmw']
+        self.assertEqual(rmw_plugin.get_name(), 'rmw')
+
+        p = rmw_plugin()
+        self.assertTrue(plugin_load_parser_correctly(rmw_plugin))
+        
+
+        mock_cliargs = {'rmw': ['cyclonedds']}
+        # TODO self.assertEqual(p.get_snippet(mock_cliargs), '')
```
codebot, line 54: Can this comment be removed, due to line 55 ?

codebot (approved):
LGTM

test/test_rmw_extension.py
tfoote: Thanks yes, it was implemented in L58/9. I'll remove it.

PR #274, opened 2024-03-18
[git extension] fix existing user_override_name key, but with None value

muritane:
When trying to `git commit` found out that the `~/.gitconfig` file was mounted to `/home/None/.gitconfig`.
The corresponding test is also added.


tfoote (approved):
Reproduced with `rocker --git --user -- ros:rolling`

The get method is only testing for defined, but now that the value has a default value it is defined but still None.

Thanks for the fix with a test to confirm coverage.

PR #277, opened 2024-04-25
Update nvidia_preamble.Dockerfile.em

pijaro:
This PR updates the Nvidia image version with the newer versions based on ubuntu 20.04 and 22.04.

This fixes some driver compatibility issues on the newest Ubuntu versions.

### DIFF
--- a/src/rocker/templates/nvidia_preamble.Dockerfile.em
+++ b/src/rocker/templates/nvidia_preamble.Dockerfile.em
@@ -1,3 +1,3 @@
 # Ubuntu 16.04 with nvidia-docker2 beta opengl support
-@{suffix = '16.04' if image_distro_version == '16.04' else '18.04'}@
+@{suffix = image_distro_version if image_distro_version in ['16.04', '18.04', '20.04', '22.04'] else '22.04'}@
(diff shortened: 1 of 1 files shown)


ahcorde (approved):

tfoote (changes requested):
Thanks this looks like this is an overdue update. However I think that this potentially breaks existing users on older platforms by forcing them to newer versions. We should make sure to have a path to have the older capabilities via an override.

So instead of just changing the template I would like to see the addition of a parameter to set `--nvidia-glvnd-version=XX.XX`   And if that's not set a policy `--nvidia-glvnd-policy=closest_lts` where we can do the logic to give the most recent LTS host similar to what you've implemented except allowing a little bit more interpolation to support not LTS versions that right now will get the 22.04 version even if it's on say 19.04. 

Then this version can be passed through to the template as a parameter.

tfoote:
@pijaro Could you try #281 to verify it works for your use case?

tfoote:
Closing this as superseded by #281

pijaro:
@tfoote Thank you for the #281, it is indeed a better approach.

tfoote:
Thanks for confirming it works for your case.

PR #281, opened 2024-08-02
Implement glvnd version option and default policy

tfoote:
This is the more full solution proposed in the review of #277


andreasBihlmaier:
Worked great in lastest WSL (based on Ubuntu 22.04) on Win11 for the `osrf/ros:iron-desktop-full` image (which is also based on Ubuntu 22.04).

Issue #241, opened 2023-08-16
Add ipc alias option


tfoote:
With `ipc=host` and `--user=host` fastrtps will work by default without needing to disable the shm transport.

From: https://robotics.stackexchange.com/a/103513/161 I've validated that talker listener work out of the box in a way that didn't before on humble.

tfoote:
Added IPC in #285

PR #285, opened 2024-08-30
Feature/added ipc flag

sachinkum0009:
- Introduced a new IPC class that extends RockerExtension to handle IPC namespace options.
- Implemented `get_name` to return the 'ipc' identifier.
- Added `get_docker_args` method to generate Docker IPC arguments based on CLI input.
- Included a `get_preamble` method that returns an empty string as no preamble is needed.

This PR enables the use of IPC-related Docker arguments in Rocker configurations and should close the issue #241.


src/rocker/extensions.py
```diff
@@ -162,6 +162,26 @@ def register_arguments(parser, defaults={}):
...
+    def get_docker_args(self, cliargs):
+        args = ''
+        ipc = cliargs.get('ipc', None)
+        args += ' --ipc %s ' % ipc
+        return args
+    
+    @staticmethod
+    def register_arguments(parser, defaults={}):
+        parser.add_argument('--ipc', default=defaults.get('ipc', 'private'),
+                            help='IPC namespace to use.')
```
tfoote, line 184:
```suggestion
                            help='IPC namespace to use. To share ipc with the host use host. More details can be found at https://docs.docker.com/reference/cli/docker/container/run/#ipc')
```

setup.py
```diff
@@ -56,6 +56,7 @@
             'hostname = rocker.extensions:Hostname',
             'name = rocker.extensions:Name',
             'network = rocker.extensions:Network',
+            'ipc = rocker.extensions:IPC',
```
tfoote, line 59: Please insert this in alphabetical order for ease of maintenance and readability.

src/rocker/extensions.py
```diff
@@ -162,6 +162,26 @@ def register_arguments(parser, defaults={}):
             default=defaults.get('network', None),
             help="What network configuration to use.")
 
+class IPC(RockerExtension):
```
tfoote, line 165: Likewise stick this in alphabetical order since it's otherwise not otherwise constrained for easier reading.

tfoote: I see Expose is in the wrong place too, but we can converge on a better layout.

tfoote (changes requested):
Thanks for the contribution, I thought that this was going to be more complicated. Overall it looks good. I have a few stylistic changes and a request for adding a full link for the options documentation which is oddly hard to get to directly. 

Can you please add a basic test too to make sure that it gets passed through correctly and we keep coverage high. You can use the network test as a good example: https://github.com/osrf/rocker/blob/73ab0d61fbcd583b89441b92640561550b634b2b/test/test_extension.py#L117-L147

setup.py
sachinkum0009: Added the basic unit test for high coverage and order the code in alphabetical order.

tfoote (approved):
Thanks for the cleanup and adding a test. This is great to add as a feature.

PR #291, opened 2024-10-09
Add support for ulimit flag

fpadula:
Hi, this PR adds support for the `ulimit` flag (https://docs.docker.com/reference/cli/docker/container/run/#ulimit). This flag is useful when running ROS 1 containers in machines with high `nofile` limits (default in Arch Linux) since not limiting this causes system crashes (please read https://answers.ros.org/question/336963/rosout-high-memory-usage/ for more info).

This PR adds:
- `Ulimit` class that extends  `RockerExtension` to add support for the `--ulimit` flag in Docker
- Checks for appropriate syntax using regex
- Descriptive message when using `help` command and when raising argument type exception (similar to the `Volume` extension)
- Basic unit tests

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -43,7 +43,7 @@ kwargs = {
         'console_scripts': [
             'rocker = rocker.cli:main',
             'detect_docker_image_os = rocker.cli:detect_image_os',
-	    ],
+        ],
         'rocker.extensions': [
             'cuda = rocker.nvidia_extension:Cuda',
             'devices = rocker.extensions:Devices',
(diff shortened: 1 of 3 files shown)


tfoote (changes requested):
Thanks for a well written plugin with good coverage. I have one request for the tests to improve the error capture if they fail. Otherwise this looks good to me.

fpadula:
> Thanks for a well written plugin with good coverage. I have one request for the tests to improve the error capture if they fail. Otherwise this looks good to me.

Thanks for the feedback! I've addressed the requests on the latest commit.

tfoote (approved):
Thanks for the iteration. It looks good to me now.

PR #292, opened 2024-10-16
Feature/remove default arguments warning for volume extension

blooop:
when running rocker this warning appears every time:
```bash
```

This sets default arguments for the volume extension to remove this warning.

It also adds type hints for the parser argument.

### DIFF
--- a/src/rocker/volume_extension.py
+++ b/src/rocker/volume_extension.py
@@ -61,7 +61,7 @@ class Volume(RockerExtension):
         return ' '.join(args)
 
     @staticmethod
-    def register_arguments(parser):
+    def register_arguments(parser, defaults: dict=None):
         parser.add_argument(Volume.ARG_ROCKER_VOLUME,
             metavar='HOST-DIR[:CONTAINER-DIR[:OPTIONS]]',
             type=str,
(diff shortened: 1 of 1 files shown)


blooop:
Oh sorry. I was not up to date with master where you already fixed this.  I guess this pr is just adding type hints now.

src/rocker/volume_extension.py
```diff
@@ -12,7 +12,7 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
-from argparse import ArgumentTypeError
+from argparse import ArgumentTypeError, ArgumentParser
```
tfoote, line 15:
```suggestion
from argparse import ArgumentParser
from argparse import ArgumentTypeError
```

tfoote (approved):
Thanks for the attention to clear the warning. Sorry we had the collision on fixing it. But adding the type annotations sounds good. 

PS would you be willing to add some of your rocker extensions to this list? https://github.com/osrf/rocker?tab=readme-ov-file#externally-maintained-extensions

blooop:
Yes, I'm happy to add my rocker extension to the list.  I was holding off because I'm still getting my head around things and didn't want to advertise them in case I needed to make any breaking changes.   In fact today I was considering making a meta-package that includes all the rocker extensions but I need to think through all the tradeoffs before I go ahead with that.

PR #293, opened 2024-10-16
Feature/external repos

blooop:
Hey,

I've updated the readme to include links to my rocker extensions.   I included template_rocker as well even though its not technically an extension, but could be helpful for other people to create their own extensions. 

I have also included links to my two rocker wrappers, but they are a bit more experimental so feel free to revert that commit.


tfoote (approved):
Thanks for submitting those!

PR #296, opened 2024-10-30
Added bashrc_extesions extension

agyoungs:
This allows you to customize the .bashrc in any container. You can achieve something similar by using the `--volume ~/.bashrc` argument, but that's not always what you want. Sometimes you want to use the base image's native .bashrc and just extend it for your use case.

You can specify a list of files/URLs.

I often use `--bashrc-extensions https://gist.githubusercontent.com/agyoungs/64f19391f864d0ea082b8a4d87a90d2b/raw/.bash-prompt.sh` as that one is super helpful for me, but this extension is generic enough to allow anyone to customize it for their use case.

### DIFF
--- a/src/rocker/bashrc_extension.py
+++ b/src/rocker/bashrc_extension.py
@@ -1,0 +1,79 @@
+import em
+import pkgutil
+import os
+import urllib3
+from rocker.core import get_user_name, ExtensionError
+from rocker.extensions import RockerExtension
+
+
+class BashrcExtensions(RockerExtension):
+
+    name = 'bashrc_extensions'
+
+    @classmethod
+    def get_name(cls):
+        return cls.name
+
+    def __init__(self):
+        self._env_subs = None
+        self.name = BashrcExtensions.get_name()
+
+    def precondition_environment(self, cli_args):
+        pass
+
+    def validate_environment(self, cli_args):
+        pass
+
+    def get_preamble(self, cli_args):
+        return ''
+    
+    def get_filename(self, bashrc_extension_file):
+        if os.path.isfile(bashrc_extension_file):
+            return os.path.join(self.name, os.path.basename(bashrc_extension_file))
+        elif bashrc_extension_file.startswith(('http://', 'https://')):
+            filename = os.path.basename(urllib3.util.url.parse_url(bashrc_extension_file).path)
+            if not filename:
+                raise ExtensionError('Bashrc extension file does not appear to have a filename: {}'.format(bashrc_extension_file))
+            return os.path.join(self.name, filename)
+        else:
+            raise ExtensionError('Bashrc extension files is not a file or URL: {}'.format(bashrc_extension_file))
+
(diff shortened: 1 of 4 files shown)


src/rocker/bashrc_extension.py
```diff
@@ -0,0 +1,79 @@
...
+
+    def get_files(self, cli_args):
+        files = {}
+        for bashrc_extension in cli_args[self.name]:
+            if os.path.isfile(bashrc_extension):
+                with open(bashrc_extension, 'r') as f:
+                    files[self.get_filename(bashrc_extension)] = f.read()
+            elif bashrc_extension.startswith(('http://', 'https://')):
+                try:
+                    response = urllib3.PoolManager().request('GET', bashrc_extension)
```
tfoote, line 49: This is a big change in scope. Fetching remote content opens up a whole extra can of worms. You can easily wget it to file just before running with barely more characters. Lets not add this level of complexity. Especially as this is a path to full remote code execution in the environment when you enter the container as we're not validating the remote content.

src/rocker/templates/bashrc_extensions_user_snippet.Dockerfile.em
```diff
@@ -0,0 +1,6 @@
+WORKDIR @home_dir
+RUN echo "\n# Source custom bashrc extensions" >> @home_dir/.bashrc
+@[for path, filename in bashrc_extension_files.items()]
+COPY @path @filename
```
tfoote, line 4: This is effectively a temporary file, it should probably go into somewhere like /tmp and not potentially override content in their home directory. (For example if someone tries to append their own local `.bashrc` it might get messy.

src/rocker/templates/bashrc_extensions_user_snippet.Dockerfile.em
```diff
@@ -0,0 +1,6 @@
+WORKDIR @home_dir
+RUN echo "\n# Source custom bashrc extensions" >> @home_dir/.bashrc
+@[for path, filename in bashrc_extension_files.items()]
```
tfoote, line 3: This is potentially adding a bunch of layers. It can at least collapse the RUN command into only one instance not one per file.

src/rocker/templates/bashrc_extensions_user_snippet.Dockerfile.em
```diff
@@ -0,0 +1,6 @@
+WORKDIR @home_dir
```
tfoote, line 1: We need to be careful about setting this WORKDIR as it sideeffects downstream snippets.

tfoote (changes requested):
Looking at this we also need tests to consider merging it. 

I'm a bit wondering if this is actually specific to the .bashrc or if it should simply be a multi argument that takes one filename in the container and then appends the contents of any of the following local paths to that filename with an append operator.  

ala `--append-file ~/.bashrc my_extension1 my_extension2 --`

Issue #298, opened 2024-11-07
Broken packages when using the cuda extension


jonazpiazu:
Looks like using the `--cuda` flag is broken, I tested for Ubuntu Focal and Ubuntu Jammy.

Running:

```bash
rocker --pull --user --x11 --nvidia auto --cuda ubuntu:jammy bash
```

Produces the error:

```bash
building > Reading package lists...
building > W: https://developer.download.nvidia.com/compute/cuda/repos/ubuntu2204/x86_64/InRelease: Key is stored in legacy trusted.gpg keyring (/etc/apt/trusted.gpg), see the DEPRECATION section in apt-key(8) for details.

building > Reading package lists...
building > Building dependency tree...
building > 
Reading state information...
building > Some packages could not be installed. This may mean that you have
requested an impossible situation or if you are using the unstable
distribution that some required packages have not yet been created
or been moved out of Incoming.
The following information may help to resolve the situation:

The following packages have unmet dependencies:
building >  nvidia-driver-565-open : Depends: libnvidia-gl-565 (= 565.57.01-0ubuntu1) but it is not installable
                          Depends: nvidia-dkms-565-open (= 565.57.01-0ubuntu1) but it is not installable
                          Depends: nvidia-kernel-common-565 (= 565.57.01-0ubuntu1) but it is not installable
                          Depends: nvidia-kernel-source-565-open (= 565.57.01-0ubuntu1) but it is not installable
                          Depends: libnvidia-compute-565 (= 565.57.01-0ubuntu1) but it is not installable
                          Depends: libnvidia-extra-565 (= 565.57.01-0ubuntu1) but it is not installable
                          Depends: nvidia-compute-utils-565 (= 565.57.01-0ubuntu1) but it is not installable
                          Depends: libnvidia-decode-565 (= 565.57.01-0ubuntu1) but it is not installable
                          Depends: libnvidia-encode-565 (= 565.57.01-0ubuntu1) but it is not installable
                          Depends: nvidia-utils-565 (= 565.57.01-0ubuntu1) but it is not installable
                          Depends: xserver-xorg-video-nvidia-565 (= 565.57.01-0ubuntu1) but it is not installable
                          Depends: libnvidia-cfg1-565 (= 565.57.01-0ubuntu1) but it is not installable
                          Depends: libnvidia-fbc1-565 (= 565.57.01-0ubuntu1) but it is not installable
                          Recommends: libnvidia-compute-565:i386 (= 565.57.01-0ubuntu1)
[...]

PR #299, opened 2024-11-07
Update cuda snippet according to nvidia official instructions

jonazpiazu:
Fixes https://github.com/osrf/rocker/issues/298


tfoote (approved):
Thanks for flagging this and providing a PR. I did a bit more updating for the new supported versions. If you have a chance to test it that would be great. I'm not going to wait for a test to merge since it's broken for all CUDA users right now.

PR #300, opened 2024-11-09
Upgrade CI usage to avoid deprecated Node version

tfoote:
checkout@v3 needs an upgrade

`The following actions use a deprecated Node.js version and will be forced to run on node20: actions/checkout@v3, actions/setup-python@v4. For more info: https://github.blog/changelog/2024-03-07-github-actions-all-actions-will-run-on-node20-instead-of-node16-by-default/`

### DIFF
--- a/.github/workflows/basic-ci.yaml
+++ b/.github/workflows/basic-ci.yaml
@@ -9,7 +9,7 @@ jobs:
       matrix:
         python-version: ['3.8', 3.x]
     steps:
-    - uses: actions/checkout@v3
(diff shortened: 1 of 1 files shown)


tfoote:
Same with setup-python@v4 -. v5

`The following actions use a deprecated Node.js version and will be forced to run on node20: actions/setup-python@v4. For more info: https://github.blog/changelog/2024-03-07-github-actions-all-actions-will-run-on-node20-instead-of-node16-by-default/`

PR #313, opened 2025-01-25
Improve the robustness of image cleanup

tfoote:
Don't crash if there's a dependent image already
And don't cleanup if the user has given it a name. The name's useless if we clear it.

### DIFF
--- a/src/rocker/core.py
+++ b/src/rocker/core.py
@@ -251,12 +251,24 @@ def docker_build(docker_client = None, output_callback = None, **kwargs):
         print("no more output and success not detected")
         return None
 
-def docker_remove_image(image_id, docker_client = None, output_callback = None, **kwargs):
+def docker_remove_image(
+        image_id,
+        docker_client = None,
+        output_callback = None,
+        fail_on_error = False,
+        force = False,
+        **kwargs):
 
     if not docker_client:
         docker_client = get_docker_client()
 
-    docker_client.remove_image(image_id)
+    try:
+        docker_client.remove_image(image_id, force=force)
+    except docker.errors.APIError as ex:
+        ## removing the image can fail if there's child images
+        if fail_on_error:
+            return False
+    return True
 
 class SIGWINCHPassthrough(object):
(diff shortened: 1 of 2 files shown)


src/rocker/core.py
```diff
@@ -251,12 +251,24 @@ def docker_build(docker_client = None, output_callback = None, **kwargs):
...
+        **kwargs):
 
     if not docker_client:
         docker_client = get_docker_client()
 
-    docker_client.remove_image(image_id)
+    try:
+        docker_client.remove_image(image_id, force=force)
+    except docker.errors.APIError as ex:
+        ## removing the image can fail if there's child images
```
nuclearsandwich, line 268: It seems a little weird not to print or log a message when catching this error. The end-user has no way of knowing whether the image was cleaned up or not, especially with the default behavior of not failing on error.

src/rocker/core.py
```diff
@@ -251,12 +251,24 @@ def docker_build(docker_client = None, output_callback = None, **kwargs):
         print("no more output and success not detected")
         return None
 
-def docker_remove_image(image_id, docker_client = None, output_callback = None, **kwargs):
+def docker_remove_image(
+        image_id,
+        docker_client = None,
+        output_callback = None,
```
nuclearsandwich, line 257: This is kind of drive-by but `output_callback` appears completely unused. Is it deprecated?

nuclearsandwich (approved):
Raised a concern about squelching the error but nothing blocking.

src/rocker/core.py
tfoote: This will return a True/False and then it warns the user based on the results: https://github.com/osrf/rocker/blob/4a6e19171765e417d211ec9a139a66987bca5645/src/rocker/core.py#L437

And the default is False so it's only suppressed if the developer asks for it to be suppressed.

src/rocker/core.py
tfoote: Yeah, this doesn't have the streaming interface so removing it makes sense.

PR #315, opened 2025-01-26
add rmw zenoh support to rmw plugin

### LINKED ISSUE #307: Add support for rmw zenoh to the RMW extension
Following up availability announced here: https://discourse.ros.org/t/rmw-zenoh-binaries-for-rolling-jazzy-and-humble/41395

The necessary steps to demo are here. 

```
FROM ros:rolling

RUN apt-get update \
    && apt-get install -qy ros-rolling-rmw-zenoh-cpp ros-rolling-demo-nodes-cpp


ENV RMW_IMPLEMENTATION=rmw_zenoh_cpp


## See the demo

#. /opt/ros/rolling/setup.bash
#ros2 run rmw_zenoh_cpp rmw_zenohd &
#ros2 run demo_nodes_cpp listener &
#ros2 run demo_nodes_cpp talker
```

tfoote:
Fixes #307

### DIFF
--- a/src/rocker/templates/rmw_snippet.Dockerfile.em
+++ b/src/rocker/templates/rmw_snippet.Dockerfile.em
@@ -4,7 +4,7 @@
 @[ if rmw ]@
 RUN \
   if [ -z "${ROS_DISTRO}" ]; then echo "ROS_DISTRO is unset cannot override RMW" ; exit 1 ; fi ;\
-  if dpkg -l @(' '.join(packages)) > /dev/null 2>&1; then \
+  if ! dpkg -l @(' '.join(packages)) | grep -q ^ii ; then \
    apt-get update \
    && DEBIAN_FRONTENT=non-interactive apt-get install -qy --no-install-recommends\
      @(' '.join(packages)) \
--- a/src/rocker/templates/rmw_snippet.Dockerfile.em
+++ b/src/rocker/templates/rmw_snippet.Dockerfile.em
@@ -4,13 +4,16 @@
 @[ if rmw ]@
 RUN \
   if [ -z "${ROS_DISTRO}" ]; then echo "ROS_DISTRO is unset cannot override RMW" ; exit 1 ; fi ;\
-  if ! dpkg -l @(' '.join(packages)) | grep -q ^ii ; then \
-   apt-get update \
-   && DEBIAN_FRONTENT=non-interactive apt-get install -qy --no-install-recommends\
-     @(' '.join(packages)) \
-   && apt-get clean ;\
+@[for package in packages]@
+  apt-get update ;\
+  if ! dpkg -l @(packages) | grep -q ^ii ; then \
+  DEBIAN_FRONTENT=non-interactive apt-get install -qy --no-install-recommends\
+    @(package) \
+  && apt-get clean ;\
   else \
-   echo "Found rmw packages @(' '.join(packages)) no need to install" ; \
-  fi
+  echo "Found rmw package @(package) no need to install" ; \
+  fi ; \
+@[end for]@
+echo "Done detecting packages for rmw"
 @[ end if ]@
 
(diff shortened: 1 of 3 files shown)


src/rocker/templates/rmw_snippet.Dockerfile.em
```diff
@@ -4,7 +4,7 @@
 @[ if rmw ]@
 RUN \
   if [ -z "${ROS_DISTRO}" ]; then echo "ROS_DISTRO is unset cannot override RMW" ; exit 1 ; fi ;\
-  if dpkg -l @(' '.join(packages)) > /dev/null 2>&1; then \
+  if ! dpkg -l @(' '.join(packages)) | grep -q ^ii ; then \
```
tfoote, line 7: This wasn't detecting if the package was installed, just if it was in the database.

ahcorde (commented):
I will leave this comment here:

zenoh requires to launch the zenoh router. Can we add a flag to launch the zenoh route too ?

tfoote:
Yeah we can do that, but lets add it separately. There's a level of coordination needed potentially to make sure that there's only one router if there's multiple containers running etc.

src/rocker/templates/rmw_snippet.Dockerfile.em
nuclearsandwich: The conditional inversion is suspect to me.

The previous condition would have installed packages if they were all listed in the apt cache. i.e. `dpkg -l ros-rolling-rmw-zenoh-cpp` -> exit 0 -> install packages.
But if the package was not in cache, it would fail and trigger the else clause.
So I think the old behavior was buggy.

However I think the new condition is going to swallow any packages that are uncached because only the last command in a pipe provides a return value.

```
# ! dpkg -l apt | grep -q ^ii
# echo $?
1
# ! dpkg -l apt notapackage | grep -q ^ii
dpkg-query: no packages found matching notapackage
# echo $?
1
```

I played around for a bit but I honestly don't know that I could produce a reasonable one-liner that would check each package in the list. The closest I came was this lil awk script to check each package in the response from `dpkg --get-selections`

```awk
$2 ~ /^install$/ { installed[$1] = 1 }
END {
	missing_packages = 0
	split(pkgin, packages)
	for (i in packages) {
		if (installed[i] != 1) {
			missing_packages++
		}
	}
	exit(missing_packages)
}
```

Used as

```
dpkg --get-selections @(' '.join(packages)) | awk -vpkgin='@(' '.join(packages))' -f missing_packages.awk
```

This would exit 0 if all packages were installed and exit >0 otherwise. You could inline the script to avoid the file it wouldn't be as readable.

tfoote: Yes, this breaks in the mulitiple packages list case. I think it's ok as long as there's only one package tested at a time. As  dpkg -l will always return ii if the package is installed. I think it might just make sense to unroll a for loop here to test each listed package separately.

tfoote: Fixed in 7cd63bf by not trying to batch detect, only detecting one.

src/rocker/templates/rmw_snippet.Dockerfile.em
```diff
@@ -4,13 +4,16 @@
 @[ if rmw ]@
 RUN \
   if [ -z "${ROS_DISTRO}" ]; then echo "ROS_DISTRO is unset cannot override RMW" ; exit 1 ; fi ;\
-  if dpkg -l @(' '.join(packages)) > /dev/null 2>&1; then \
-   apt-get update \
-   && DEBIAN_FRONTENT=non-interactive apt-get install -qy --no-install-recommends\
-     @(' '.join(packages)) \
-   && apt-get clean ;\
+@[for package in packages]@
+  apt-get update ;\
```
nuclearsandwich, line 8: Minor nit: doing the apt-get update and apt-get clean as setup and teardown to the loop would cut down on timing. It's not important while the loop size is usually 1.

nuclearsandwich (approved):

PR #317, opened 2025-01-30
Enable redirecting non-interactive run console output to file

tfoote:


src/rocker/core.py
```diff
@@ -399,10 +401,13 @@ def run(self, command='', **kwargs):
...
-                print(cmd)
-                p = subprocess.run(shlex.split(cmd), check=True, stderr=subprocess.STDOUT)
-                return p.returncode
+                with open(console_output_file, 'a') if console_output_file else nullcontext() as consoleout_fh:
+                    if console_output_file:
+                        print(f"Logging output to file {console_output_file}")
+                    print("Executing command: ")
+                    print(cmd)
+                    p = subprocess.run(shlex.split(cmd), check=True, stdout=consoleout_fh if console_output_file else None, stderr=subprocess.STDOUT)
+                    return p.returncode
```
nuclearsandwich, line 410: This was my first time seeing nullcontext. I'm not sure if I love it but I juggled a few different methods for handling this and all were more error prone than using the context manager unconditionally.

nuclearsandwich (approved):

src/rocker/core.py
tfoote: Yeah, i did a bunch of searching on it and it eventually seemed like the best way to not have two branches of almost duplicate code.

PR #319, opened 2025-02-08
361: Feature cuda skip install

Ravencrest-01:
This pull request enhances the CUDA extension to skip the installation step if CUDA is already present on the system. It adds a check to detect whether CUDA is installed using the nvidia-smi command, ensuring that redundant installations are avoided.

**Key Changes**

1. Added function `is_cuda_installed()` to check if CUDA is already installed.
2. Modified `get_snippet()` method to: Skip generating the Dockerfile if CUDA is detected.

### DIFF
--- a/src/rocker/cli.py
+++ b/src/rocker/cli.py
@@ -37,6 +37,7 @@ def main():
     parser.add_argument('--nocleanup', action='store_true', help='do not remove the docker container when stopped')
     parser.add_argument('--persist-image', action='store_true', help='do not remove the docker image when stopped', default=False) #TODO(tfoote) Add a name to it if persisting
     parser.add_argument('--pull', action='store_true')
+    parser.add_argument('--dive', action='store_true', help='Run Dive for introspection after building the image') # 310
     parser.add_argument('--version', action='version',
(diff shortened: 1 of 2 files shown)


Ravencrest-01:
Hi @tfoote,

I’ve implemented the enhancement to the CUDA extension to skip the installation if CUDA is already present on the system. This change introduces a check using the `nvidia-smi` command and ensures that redundant installations are avoided, improving efficiency.

I’ve tested the feature under the following conditions:
- **With CUDA installed:** The system correctly skips installation and logs the message: `"CUDA is already installed. Skipping installation."`
- **Without CUDA installed:** The extension proceeds with the installation as expected.

Looking forward to your feedback! Please let me know if there’s anything else I can improve.

Best regards,  
Ritik kumar

tfoote (changes requested):
Thanks for looking at this, however this won't do what is needed.  This is checking the host, not the container or the presence of cuda.

tfoote:
This needs major changes before it will be ready so I'm going to push this back to draft status. It can be rereviewed when the additional functionality has been implemented.

PR #320, opened 2025-02-14
add detach plugin

YuqiHuai:
When rocker launches a docker container, there was no option to run it in detached mode. I added a plugin and a test case for this plugin that allows rocker to run container in detached mode.

All test cases in `test_extension.py` passes.
```
❯ python3 -m pytest test/test_extension.py
============================================= test session starts ==============================================
platform linux -- Python 3.13.1, pytest-8.3.4, pluggy-1.5.0
rootdir: /home/yuqi/ResearchWorkspace/rocker
configfile: setup.cfg
plugins: ament-flake8-0.12.11, ament-copyright-0.12.11, launch-testing-ros-0.19.7, ament-pep257-0.12.11, ament-lint-0.12.11, ament-xmllint-0.12.11, launch-testing-1.0.6, cov-6.0.0
collected 19 items                                                                                             

test/test_extension.py ...................                                                               [100%]

============================================= 19 passed in 23.75s ==============================================
~/ResearchWorkspace/rocker main ❯    
```

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -46,6 +46,7 @@ kwargs = {
         ],
         'rocker.extensions': [
             'cuda = rocker.nvidia_extension:Cuda',
+            'detach = rocker.extensions:Detach',
             'devices = rocker.extensions:Devices',
             'dev_helpers = rocker.extensions:DevHelpers',
             'env = rocker.extensions:Environment',
@@ -56,7 +57,6 @@ kwargs = {
             'hostname = rocker.extensions:Hostname',
             'ipc = rocker.extensions:Ipc',
             'name = rocker.extensions:Name',
-            'detach = rocker.extensions:Detach',
             'network = rocker.extensions:Network',
             'nvidia = rocker.nvidia_extension:Nvidia',
             'port = rocker.extensions:Port',
--- a/setup.py
+++ b/setup.py
@@ -56,6 +56,7 @@ kwargs = {
             'hostname = rocker.extensions:Hostname',
             'ipc = rocker.extensions:Ipc',
             'name = rocker.extensions:Name',
+            'detach = rocker.extensions:Detach',
             'network = rocker.extensions:Network',
             'nvidia = rocker.nvidia_extension:Nvidia',
             'port = rocker.extensions:Port',
--- a/src/rocker/extensions.py
+++ b/src/rocker/extensions.py
@@ -32,6 +32,32 @@ def name_to_argument(name):
 
 from .core import RockerExtension
 
+class Detach(RockerExtension):
+    @staticmethod
+    def get_name():
+        return 'detach'
+    
+    def __init__(self):
+        self.name = Detach.get_name()
+    
+    def get_docker_args(self, cliargs):
+        args = ''
+        detach = cliargs.get('detach', False)
+        if detach:
+            args += ' --detach'
+        return args
+    
+    @staticmethod
+    def register_arguments(parser, defaults):
+        parser.add_argument(
+            '-d',
+            '--detach',
+            action='store_true',
+            default=defaults.get('detach', False),
+            help='Run the container in the background.'
+        )
+
+
 class Devices(RockerExtension):
     @staticmethod
     def get_name():
@@ -186,29 +212,6 @@ class Name(RockerExtension):
         parser.add_argument('--name', default=defaults.get('name', ''),
                             help='Name of the container.')
 
-class Detach(RockerExtension):
-    @staticmethod
-    def get_name():
-        return 'detach'
-    
-    def __init__(self):
-        self.name = Detach.get_name()
-    
-    def get_docker_args(self, cliargs):
-        args = ''
-        detach = cliargs.get('detach', False)
-        if detach:
-            args += ' --detach'
-        return args
-    
-    @staticmethod
-    def register_arguments(parser, defaults):
-        parser.add_argument(
-            '--detach',
-            action='store_true',
-            default=defaults.get('detach', False),
-            help='Run the container in the background.'
-        )
 
 class Network(RockerExtension):
     @staticmethod
--- a/src/rocker/extensions.py
+++ b/src/rocker/extensions.py
@@ -186,6 +186,29 @@ class Name(RockerExtension):
         parser.add_argument('--name', default=defaults.get('name', ''),
                             help='Name of the container.')
 
+class Detach(RockerExtension):
+    @staticmethod
+    def get_name():
+        return 'detach'
+    
+    def __init__(self):
+        self.name = Detach.get_name()
+    
+    def get_docker_args(self, cliargs):
+        args = ''
+        detach = cliargs.get('detach', False)
+        if detach:
+            args += ' --detach'
+        return args
+    
+    @staticmethod
+    def register_arguments(parser, defaults):
+        parser.add_argument(
+            '--detach',
+            action='store_true',
+            default=defaults.get('detach', False),
+            help='Run the container in the background.'
+        )
 
 class Network(RockerExtension):
     @staticmethod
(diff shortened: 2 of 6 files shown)


src/rocker/extensions.py
```diff
@@ -186,6 +186,29 @@ def register_arguments(parser, defaults):
...
+        args = ''
+        detach = cliargs.get('detach', False)
+        if detach:
+            args += ' --detach'
+        return args
+    
+    @staticmethod
+    def register_arguments(parser, defaults):
+        parser.add_argument(
+            '--detach',
```
tfoote, line 207: It would be good to leverage the same command line arguments shorthand as docker " -d, --detach"

```suggestion
            '-d',
            '--detach',
```

src/rocker/extensions.py
```diff
@@ -186,6 +186,29 @@ def register_arguments(parser, defaults):
         parser.add_argument('--name', default=defaults.get('name', ''),
                             help='Name of the container.')
 
+class Detach(RockerExtension):
```
tfoote, line 189: Please put this in alphabetical order

setup.py
```diff
@@ -56,6 +56,7 @@
             'hostname = rocker.extensions:Hostname',
             'ipc = rocker.extensions:Ipc',
             'name = rocker.extensions:Name',
+            'detach = rocker.extensions:Detach',
```
tfoote, line 59: Please put this in alphabetical order.

test/test_extension.py
```diff
@@ -308,6 +308,39 @@ def test_name_extension(self):
         args = p.get_docker_args(mock_cliargs)
         self.assertTrue('--hostname docker-hostname' in args)
 
+class DetachExtensionTest(unittest.TestCase):
```
tfoote, line 311: Please put this in alphabetical order too.

tfoote (changes requested):
Thanks for this submission, it looks great to be able to provide this option. I had a few stylistic requests. 

Can you explain a bit about how you're using this? I want to connect this to the mode setting and potentially force NON-INTERACTIVE in this case, or error if the mode is set explicitly to interactive.

tfoote:
Thanks for reordering, I have tweaked the command line parsing to handle mutual exclusion of --detach and interactive mode: https://github.com/YuqiHuai/rocker/pull/1

Can you validate that works for you and then I think this will be good to go.

YuqiHuai:
@tfoote Thanks for making these changes! Unfortunately, with those changes, detach no longer works properly for my purpose. Setting --detach seems to automatically activate the non-interactive mode, and the container I am trying to get running doesn't run in the background when the mode is set to non-interactive.

I am going to test a bit more to find the root cause.

tfoote:
Interesting, I guess that maybe your program is detecting the pseudo tty of the pexpect interface and doing something different. I've been testing with:

`rocker  --mode interactive ros:jazzy sleep 10` Run normally
`rocker --detach ros:jazzy sleep 10` Detach and run in the background for 10 seconds (visible via docker ps)
`echo foo | rocker ros:jazzy sleep 10`  Trigger no-tty fallback

Errors:
`rocker  --mode interactive --detach ros:jazzy sleep 10`  # Interactive and detach are not compatible
`echo foo | rocker  --mode interactive ros:jazzy sleep 10 ` # no-tty and interactive are not compatible

I guess it would be possible to continue to keep the pexpect invocation but this seems like an odd edge case. And the intent of the interactive environment is to connect the users' keyboard. And when you're running headless it seems like if the executable is running or expecting user inputs that's not going to get very far.

YuqiHuai:
@tfoote I created a test image
```Dockerfile
FROM ubuntu:18.04
CMD ["/bin/bash"]
```
and tried to run different docker commands to see how I can achieve my goal (launching a container without necessarily entering it). It turns out I need flags `-itd` (both interactive and detached). I am not sure, from a design perspective of rocker, whether "interactive" and "detach" should be mutually exclusive since docker actually allows it. (See example https://github.com/ApolloAuto/apollo/blob/463fb82f9e979d02dcb25044e60931293ab2dba0/docker/scripts/dev_start.sh#L384)

After removing your changes of setting mode to non-interactive when detach is set, and adding code to set default to interactive, things are working as how I expect it to run.

YuqiHuai:
@tfoote I Just saw your comment after I posted the previous comment. I think the use case of -it and -d is different for everyone. Perhaps some users are trying to run an executable in a container and they don't need to keep the container up and running after the executable finishes (e.g., `docker run --rm --detach ros:jazzy sleep 10`). For my purpose, I want to keep the container up and running, so there needs to be a detached interactive session with the `-its` flag (e.g., `docker run --rm -itd XXX /bin/bash`).

tfoote:
None of these cases keep the container running after the executable finishes. Some applications will finish after stdin is disconnected, and the only common example I know of this is interpreters like bash. For example `/bin/bash` without `-i` will terminate when the bash script reaches the end of the inputs which is nothing, so it will immediately end.  

I found a good overview of the distinctions here: https://stackoverflow.com/questions/47753047/docker-detached-and-interactive

I see from your linked `dev_start.sh` what you're relying on is not the executable, but that you want the dev envrionment to be just running in the background. You should change the last line from /bin/bash to `sleep infinity` which will mean that it just waits. Right now you're relying on /bin/bash waiting infinitely on stdin which is good if you want to interact, but in a non-interactive case is not a good way to express that you want to wait forever. 

Usually the end of your docker command would be run my "service" aka a web server or the like. And the setup beforehand is to make sure all the dependencies are running in the container. 

To stop it from immediately ending you can pass a command like `sleep infinity`

`rocker --nvidia --privileged --x11 --user --mode non-interactive -- ghcr.io/autowarefoundation/autoware:20240903-devel-cuda-amd64 sleep infinity` which will pass sleep infinity to the bash instance to keep it alive. 

This is the same as the default commands in `ros:jazzy` if you just run `rocker -d ros:jazzy` it will exit immediately because bash finishes. However `rocker -d ros:jazzy sleep infinity` will run until it's killed. 

I'm downloading your image but it's going to be a while to finish and be able to reproduce.

YuqiHuai:
@tfoote Thanks for taking a close look at the specific use case! I think your characterization of "I want the dev environment to be just running in the background" is accurate. Your example of using non-interactive mode with an additional "sleep infinity" command looks like it will achieve my goal but block the main terminal at the same time.

I think we already understood the situation pretty well (i.e., using `-itd` and `/bin/bash` to keep a container running in the background without doing anything). You might not need to download the image. They are indeed large :'(
 
Back to this PR, I still believe having a detach flag helps, but making it mutually exclusive with interactive might be problematic because it is supported by docker. Do you think this detach plugin still contributes to rocker without making it mutually exclusive w.r.t. mode?

tfoote:
The detach capability is something that I have other use cases that could be leveraged for example: https://github.com/opencv/bpc 

However I think that detached and interactive in the rocker sense **are** mutually exclusive, you cannot interact with a detached instance. There's a question of if the program things that there's still a stdin, which docker calls "interactive" this is an interesting distinction that I'm not sure I agree with. 

>  Your example of using non-interactive mode with an additional "sleep infinity" command looks like it will achieve my goal but block the main terminal at the same time.

The expectation of a non-interactive process is that it will run to completion. And if you want it to run in the background detached it does the same. We could add the -i --interactive option to do the stdin mounting separately from the --mode mechanism. 

Looking a little deeper it would be great to also enable rocker to keep track of the container and enable `attach` as well to the container, either reconnecting to an rocker interactive session or to pipe in a specific input to the detached instance.

YuqiHuai:
That all sounds good. I think it is ok for rocker and docker to have different interpretations of interactive. I tested my use case and found running non-interactive, detach, sleep infinity achieves my goal of keeping a container up and running in the back. When I re-attach and run `ps -ef`, I get
```
UID          PID    PPID  C STIME TTY          TIME CMD
yuqi           1       0  0 17:47 ?        00:00:00 sleep infinity
...
```

Thank you for your time and consideration! I have merged your changes to the branch.

tfoote (approved):
Great, I'm glad it's working for you now. I'll bring this in as is for now. But I'll open a feature request for extending the interactivity to potentially detached processes.

YuqiHuai:
Thanks again for your time and consideration! I learned a lot about rocker and docker from our conversation.

PR #324, opened 2025-03-27
Add config argument parsing capability

agyoungs:
Use the `ConfigArgParse` library to support passing in rocker configurations with a config (toml, ini, yaml, etc) file.

Currently the `ConfigArgParse` library needs some updates, so I've put this work on hold, but as those updates are made and I make progress on this branch I figured I'd create a placeholder here.


tfoote:
Thanks for the update and draft here for future visibilty. This looks like a great way to make it happen, but those limitations you flagged are pretty much blockers.

PR #327, opened 2025-05-10
Add workdir support

### LINKED ISSUE #249: Support --workdir
While running docker, we can assign `--workdir` to override default WORKDIR in Dockerfile.
Is it possible to support the argument in rocker command?
https://docs.docker.com/engine/reference/run/#workdir

raultapia:
Closes #249

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -68,6 +68,7 @@ kwargs = {
             'ulimit = rocker.ulimit_extension:Ulimit',
             'user = rocker.extensions:User',
             'volume = rocker.volume_extension:Volume',
+            'workdir = rocker.extensions:WorkDir',
             'x11 = rocker.nvidia_extension:X11',
         ]
     },
--- a/src/rocker/extensions.py
+++ b/src/rocker/extensions.py
@@ -302,6 +302,24 @@ class PulseAudio(RockerExtension):
             help="mount pulse audio devices")
 
 
+class WorkDir(RockerExtension):
+    @staticmethod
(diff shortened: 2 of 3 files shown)


raultapia:
I've implemented the hierarchy you proposed with some minor changes. The current status is:
- If executed without `--workdir`, defaults to home directory.
- If executed with `--workdir`, the current directory is set as `WORKDIR` from `get_user_snippet`.
- If executed with `--workdir=path`, the directory <path> is set and it overrides the `WORKDIR` from `get_user_snippet`.

tfoote:
Thanks for iterating this is making progress, but it isn't dealing withthe workdir interactions with the user extension: https://github.com/osrf/rocker/blob/ade980316020c2e3e82412081bb656c73bd85556/src/rocker/templates/user_snippet.Dockerfile.em#L38 

And the workdir will need to be the last snippet to be effective for setting the workdir for the user only, and now potentially effecting the other extensions if it gets evaluated first. I think that it may make more sense to do this as a core capability versus as an extension because of that and the need for the interaction with the mounting and user mechanisms.

tfoote (changes requested):
Please see my comments in the main thread. This needs to provide sequencing protection and interact with the user and mount extensions to give the expected potential working directories. 

somewhere like https://github.com/osrf/rocker/blob/ade980316020c2e3e82412081bb656c73bd85556/src/rocker/core.py#L485

Or else we need to add a sequencing support to extensions.

PR #359, opened 2026-05-04
fix: handle User extension test when invoking user has no groups (#339)

arnav03op:
Fixes #339 

**Description of Changes:**
The `test_user_extension` unit test previously failed for host users who are not explicitly members of any supplementary groups. The test was unconditionally asserting the presence of the `usermod -aG` command in the generated snippet, which `rocker` correctly omits when there are no groups to add.

This PR introduces a dynamic check using the `grp` module to fetch the invoking user's actual group memberships. Conditional logic was added so that the test only expects `usermod -aG` and `user-preserve-group-permissive Enabled` in the snippet if the user actually belongs to secondary groups. If the user has no supplementary groups, it properly asserts `False`.

**Testing Performed:**
Ran `pytest test/test_extension.py` locally to verify that the tests now pass cleanly in environments where the host user lacks secondary group memberships.

### DIFF
--- a/.github/workflows/basic-ci.yaml
+++ b/.github/workflows/basic-ci.yaml
@@ -26,7 +26,7 @@ jobs:
       with:
         run: python -m pytest -s -v --cov=rocker -m "not nvidia"
     - name: Upload coverage to Codecov
-      uses: codecov/codecov-action@v4
+      uses: codecov/codecov-action@v6
       with:
         use_oidc: true
   cli_smoke_tests:
--- a/src/rocker/extensions.py
+++ b/src/rocker/extensions.py
@@ -357,7 +357,10 @@ class User(RockerExtension):
                     print('Warning skipping groups %s because they do not exist on the host.' % unmatched_groups)
                 substitutions['user_groups'] = ' '.join(['{};{}'.format(g.gr_name, g.gr_gid) for g in matched_groups])
             else:
-                substitutions['user_groups'] = ' '.join(['{};{}'.format(g.gr_name, g.gr_gid) for g in all_groups if substitutions['name'] in g.gr_mem])
+                matched_groups = [g for g in all_groups if substitutions['name'] in g.gr_mem]
+                if not matched_groups:
+                    print('User %s is not a member of any supplementary groups, skipping group preservation.' % substitutions['name'])
+                substitutions['user_groups'] = ' '.join(['{};{}'.format(g.gr_name, g.gr_gid) for g in matched_groups])
         else:
             substitutions['user_groups'] = ''
         substitutions['user_preserve_groups_permissive'] = True if 'user_preserve_groups_permissive' in cliargs and cliargs['user_preserve_groups_permissive'] else False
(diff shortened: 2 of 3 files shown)


arnav03op:
I raised the PR, please review it and if any issues or want me to change anything or in case of any errors just let me know , I'll see it

tfoote (changes requested):
This looks like a potential solution. I haven't looked too deep at the technical solution yet as I think this should be extended structurally first. 

 I don't see a change in the code that correlates with the test changes. We should fix the issue for users not just the tests. And instead of just adding conditional logic in the tests we should mock the case to check that both paths through the code work, and not just the one present on the host system for the tests.

arnav03op:
> This looks like a potential solution. I haven't looked too deep at the technical solution yet as I think this should be extended structurally first.
> 
> I don't see a change in the code that correlates with the test changes. We should fix the issue for users not just the tests. And instead of just adding conditional logic in the tests we should mock the case to check that both paths through the code work, and not just the one present on the host system for the tests.


Thanks for the feedback ! I got it
1-Fix the actual code so it handles users with no groups .
2-Update the tests to mock both cases (with groups and without groups) 
   so the tests don't depend on the host system's user configuration

Is that what you want me to do, 
I'll work on both and update the PR shortly.

arnav03op:
Hi @tfoote , thanks for the feedback and I've updated the PR

1-Production code fix (src/rocker/extensions.py) : In User.get_snippet(), the auto-discovery path (when user_preserve_groups=[]) now explicitly handles the case where the user has no supplementary groups. The one-liner was split for clarity and an informational message is printed when no groups are found, making the no-groups path visible rather than silently producing an empty string.

2-Mocked tests (test/test_extension.py): Replaced all host-dependent conditional assertions with unittest.mock.patch on rocker.extensions.grp.getgrall. The test now deterministically covers both code paths regardless of the host system's user configuration.
-With groups: mocks a group containing the user -> asserts usermod -aG is present
-Without groups: mocks a group where the user is not a member -> asserts usermod -aG is absent
-Same for the permissive mode path
-Explicit group list test is also mocked for consistency

This ensures the tests no longer depend on the host user's actual group memberships and both paths are always exercised.

Plz let me know if you'd like any further changes!

arnav03op:
Hey @tfoote 
i want to discuss regarding the CI failure is that the failing check is the "Upload coverage to Codecov" step (codecov-action@v4 with use_oidc: true), which can't access "ACTIONS_ID_TOKEN_REQUEST_URL" on fork PRs. The actual test suite ("Run headless tests") passed successfully. This appears to be a pre-existing CI configuration issue unrelated to my changes. Am I right on this??
Please let me know if you'd like any further changes! I'll gladly look into it

tfoote (approved):
Thanks for the iteration this looks better with the fix for the usage and then good coverage on the tests. 

You're right that codecov@v4 had an issue submitting from forks using OIDC that was fixed so I bumped it up to v6 and the upload worked.

arnav03op:
Thank you @tfoote for the guidance in solving the issue.

PR #360, opened 2026-07-15
Add ubuntu 26.04 distro to nvidia extension

tpoignonec:
On Ubuntu 26.04, can't use the `--nvidia` or `--cuda` extension.
This PR just adds Ubuntu 26.04 to the list of allowed distros.


tfoote (approved):
Thanks for contributing this. It looks good to me.

PR #362, opened 2026-08-24
speed up os detection step by bind-mounting distro-detect

jonazpiazu:
I am running rocker on a custom "fat" image (by "fat" I mean around 16GB). I see that the running times for the distro-detect step are really high. Here some comparison for the vanilla ubuntu:noble image, and for my custom image:

```bash
time rocker --nvidia --mode non-interactive ubuntu:noble true
rocker --nvidia --mode non-interactive ubuntu:noble true  0,16s user 0,04s system 1% cpu 12,641 total
```

```bash
time rocker --nvidia --mode non-interactive custom-fat:jazzy true
rocker --nvidia --mode non-interactive custom-fat:jazzy true  0,15s user 0,04s system 0% cpu 1:21,81 total
```

Even a second run would give me comparable running times for the custom fat image.

I see that most of the time is spent here:

```
Step 4/6 : COPY --from=detector /go/distro-detect/distro-detect /tmp/detect_os
 ---> f3bac151e657
Step 5/6 : ENTRYPOINT [ "/tmp/detect_os", "-format", "json-one-line" ]
 ---> Running in d508eaf23ac6
 ---> Removed intermediate container d508eaf23ac6
 ---> 57b4df1bc7fd
Step 6/6 : CMD [ "" ]
 ---> Running in 47f7c05b140c
```

My understanding is that each of those lines in the generated Dockerfile is creating a new Docker layer, and thus copying/committing the 16GBs of the original image.

So the PR introduces two main changes:
- it generates a small tagged images for `distro-detect` which can be reused among runs, instead of compiling it on every detect_os() call
- instead of using the "FROM" to copy the `detect_os`, it bind-mounts the location of the `distro-detect` binary

First run (distro-detect image not cached):

```
time rocker --nvidia --mode non-interactive custom-fat:jazzy true
rocker --nvidia --mode non-interactive custom-fat:jazzy true  0,15s user 0,03s system 1% cpu 11,632 total
```

Second run:

```
time rocker --nvidia --mode non-interactive custom-fat:jazzy true
rocker --nvidia --mode non-interactive custom-fat:jazzy true  0,16s user 0,03s system 15% cpu 1,234 total
```

Disclaimer: the changes were supported by using Cursor - not sure about the specific model used as it was set in "Auto Balance"


tfoote (changes requested):
Overall I think that this approach makes a lot more sense. We definitely don't need to be building the images just to check it. And then there's also potentially a bunch of extra images too. 

I think that we can leverage the runtime image mount mechanism: https://docs.docker.com/engine/storage/image-mounts/ instead of needing to extract and inject the detector. 

Overall as feedback this has a ton of fragile helper functions that have a lot of heuristics and adds a lot of lines of code. It seems like a lot of these are partially reimplementing pexpect functionality in a more fragile way.

It would be great to see if this could be implemented much more minimally replacing the image building with a container building and running doing the image mount. All of the logic for checking the outputs should be able to be unchanged.

PR #363, opened 2026-09-17
Add ability to override home directory when using User extension

miguelprada:
I started to have problems with a `--home` + `--user` based workflow on my corporate computer where the user and group names contain special symbols. The version of `groupadd` shipped with Ubuntu 26.04 doesn't accept using special characters in the group matching my user name created automatically [here](https://github.com/osrf/rocker/blob/00bf38f685ffab5d5a250e484bd13809161cb08a/src/rocker/templates/user_snippet.Dockerfile.em#L17).

I can work around this by changing the user name inside the container with `--user-override-name`, but this [also changes the `dir` substitution](https://github.com/osrf/rocker/blob/00bf38f685ffab5d5a250e484bd13809161cb08a/src/rocker/extensions.py#L347), making the home directory of the user in the container different than the one mounted by `--home`.

By adding an optional `--user-override-home` I can set the `dir` substitution, and therefore my home in the container, back to my host user's home directory.


tfoote (commented):
Thanks adding this option makes sense. That's a tough one with special characters in those fields. On reading it it looks good. 

But I see an error in CI

E       AssertionError: '-v /tmp/tmpgzd_36vq:/home/otherdir/.gitconfig' not found in ' -v /etc/gitconfig:/etc/gitconfig:ro -v /tmp/tmpgzd_36vq:/root/.gitconfig:ro'

Which looks like a root vs user problem. I don't yet see how that comes from this change. It's possible that something in the underlying runner changed.

miguelprada:
Turns out I had introduced a typo in the modified git extension logic.

Must admit I didn't originally run the test locally because in my setup there's yet another quirk, i.e. my user being `some_id@some_domain` whereas home directory is `/home/some_domain/some_id`, breaking the `/home/{username}` assumption.

Apologies if this made you loose your time... :sweat:

I can still see some failing test, but that does indeed seem to be unrelated to this, isn't it?

